PR #6, opened 2025-09-15
feat: Add Rancher commands

amstuta:
Add commands for Rancher product:
- Creation / deletion of a service
- Fetch reference data: available versions and plans

### DIFF
--- a/internal/display/display.go
+++ b/internal/display/display.go
@@ -35,6 +35,7 @@ func renderCustomFormat(value any, format string) {
 
 	switch reflect.TypeOf(value).Kind() {
 	case reflect.Slice:
+		var output strings.Builder
 		for _, val := range value.([]map[string]any) {
 			out, err := ev(context.Background(), val)
 			if err != nil {
@@ -45,8 +46,11 @@ func renderCustomFormat(value any, format string) {
 			if err != nil {
 				ExitError("error marshalling result")
 			}
-			Output(string(outBytes))
+			output.Write(outBytes)
+			output.WriteString("\n")
 		}
+		ResultString = output.String()
+		fmt.Print(ResultString)
 	default:
 		out, err := ev(context.Background(), value)
 		if err != nil {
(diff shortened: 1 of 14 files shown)


internal/display/display.go
```diff
@@ -35,6 +35,7 @@ func renderCustomFormat(value any, format string) {
 
 	switch reflect.TypeOf(value).Kind() {
 	case reflect.Slice:
+		var output string
```
maxatome: use `strings.Builder` instead.

internal/display/display.go
```diff
@@ -150,7 +153,7 @@ func RenderTable(values []map[string]any, columnsToDisplay []string, outputForma
 		Headers(columnsTitles...).
 		Rows(rows...)
 
-	Output(t.String() + "\n💡 Use option --json or --yaml to get the raw output with all information")
+	Output("%s", t.String()+"\n💡 Use option --json or --yaml to get the raw output with all information")
```
maxatome:
```suggestion
	Output("%s%s", t, "\n💡 Use option --json or --yaml to get the raw output with all information")
```

internal/display/display.go
```diff
@@ -197,7 +200,7 @@ func RenderConfigTable(cfg *ini.File) {
 		Headers(columns...).
 		Rows(rows...)
 
-	Output(t.String())
+	Output("%s", t.String())
```
maxatome:
```suggestion
	Output("%s", t)
```

internal/display/display.go
```diff
@@ -206,7 +209,7 @@ func prettyPrintJSON(value any) error {
 		return err
 	}
 
-	Output(string(bytesOut))
+	Output("%s", string(bytesOut))
```
maxatome:
```suggestion
	Output("%s", bytesOut)
```

internal/display/display.go
```diff
@@ -217,7 +220,7 @@ func prettyPrintYAML(value any) error {
...
```
maxatome:
```suggestion
	Output("%s", bytesOut)
```

internal/display/display.go
```diff
@@ -296,7 +299,7 @@ func ExitWarning(message string, params ...any) {
 	os.Exit(0)
 }
 
-func Output(message string) {
-	fmt.Println(message)
-	ResultString = message
+func Output(message string, params ...any) {
```
maxatome: Could be named `Outputf` instead?

internal/display/display.go
```diff
@@ -45,8 +46,11 @@ func renderCustomFormat(value any, format string) {
 			if err != nil {
 				ExitError("error marshalling result")
 			}
-			Output(string(outBytes))
+			output.Write(outBytes)
+			output.WriteString("\n")
```
maxatome, line 50: just as a note, could be:
```suggestion
			output.WriteByte('\n')
```

PR #41, opened 2025-09-29
feat(me): add me command to get information about the current user

mJehanno:
# Description

Add a new `me` command to retrieve current user information.
It also fix an annoying behaviour on the makefile where `Make` would also look for folder in `$PATH` containing `go` instead of just looking for binaries.

## Type of change

# Checklist:

- [X] I updated the documentation by running `make doc`
- [X] I ran `go mod tidy`


internal/services/me/templates/me.tmpl
```diff
@@ -0,0 +1,15 @@
+👾 Me
+=======
+
+## Contact Informations
```
deathiop:
```suggestion
## Contact Information
```

internal/display/template.go
```diff
@@ -27,6 +30,8 @@ var (
 			return strings.Join(strElements, ", ")
 		},
 		"formatByteSize": formatByteSize,
+		"toTitle":        cases.Title(language.English).String,
+		"toUpper":        strings.ToUpper,
```
deathiop, line 34: If you're going this way, add `toLower`, it doesn't make sense not to have both

internal/cmd/me.go
```diff
@@ -0,0 +1,21 @@
...
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	meCmd := &cobra.Command{
+		Use:   "me",
+		Short: "Retrieve your personal information",
+	}
+
+	meCmd.AddCommand(&cobra.Command{
```
amstuta: could you move this command under `account` cmd (like `ovhcloud account get`) ?

internal/services/me/templates/me.tmpl
```diff
@@ -0,0 +1,15 @@
+👾 Me
```
amstuta:
```suggestion
👾 Account details
```

Makefile
```diff
@@ -13,18 +13,18 @@ wasm:
 	GOOS=js GOARCH=wasm ${BUILD_CMD} -ldflags "${LD_FLAGS}" -o ovhcloud.wasm ./cmd/ovhcloud
 
 test:
-	go test -v ./...
+	go test -v ./... ;
```
amstuta: I don't get this change, this is a shell line so adding a `;` should not change anything. Could you explain the change in behavior you expect ?

amstuta (changes requested):
Thanks for your contribution @mJehanno :)

Makefile
mJehanno: > It also fix an annoying behaviour on the makefile where Make would also look for folder in $PATH containing go instead of just looking for binaries.

When trying to run make command, I had a "permission denied" issue on `go`. It appears that make will look in the path for anything that contains `go` even if its a folder/subfolder, what is needed here is looking for the binary that is named `go` (https://stackoverflow.com/questions/73908737/permission-denied-when-running-go-from-makefile)

internal/display/template.go
mJehanno: I will but my thought here was to add only things that are used (kinda taking the YAGNI approach on that)

Makefile
amstuta: Honestly this seems like an unconventional setup to have subdirectories in directories that you reference in $PATH (even the SO you linked says `You should check your PATH and make sure to remove any directories that contain such subdirectories` as first solution).

We would prefer to avoid running a shell just to prevent this issue.

doc/ovhcloud.md
```diff
@@ -95,15 +95,16 @@ Below is the full list of primary sub‑commands available at the time of writin
...
 * [ovhcloud iam](ovhcloud_iam.md)	 - Manage IAM resources, permissions and policies
-* [ovhcloud ip](ovhcloud_ip.md)	 - Retrieve information and manage your IP services
-* [ovhcloud iploadbalancing](ovhcloud_iploadbalancing.md)	 - Retrieve information and manage your IP LoadBalancing services
-* [ovhcloud ldp](ovhcloud_ldp.md)	 - Retrieve information and manage your LDP (Logs Data Platform) services
+* [ovhcloud ip](ovhcloud_ip.md)	 - Retrieve information and manage your Ip services
+* [ovhcloud iploadbalancing](ovhcloud_iploadbalancing.md)	 - Retrieve information and manage your IpLoadbalancing services
+* [ovhcloud ldp](ovhcloud_ldp.md)	 - Retrieve information and manage your Ldp services
 * [ovhcloud location](ovhcloud_location.md)	 - Retrieve information and manage your Location services
 * [ovhcloud login](ovhcloud_login.md)	 - Login to your OVHcloud account to create API credentials
+* [ovhcloud me](ovhcloud_me.md)	 - Retrieve your personal information
```
amstuta: to remove as command has moved under `account`

internal/services/me/me.go
```diff
@@ -0,0 +1,17 @@
+package me
```
amstuta: can you also move the code and templates under `account/` directory ?

PR #45, opened 2025-09-30
Feat/reorder inst dirs preference order

### LINKED ISSUE #42: [Bug]: Install dir is not the one being advertised

### Current Behavior

running the install script I got the following output:

```
❯ curl -fsSL https://raw.githubusercontent.com/ovh/ovhcloud-cli/main/install.sh | sh
ovh/ovhcloud-cli info checking GitHub for latest tag
ovh/ovhcloud-cli info found version: 0.5.0 for v0.5.0/Darwin/arm64
ovh/ovhcloud-cli info installed ./bin/ovhcloud
```

However, the CLI was not installed in `.bin`folder but in the `bin`folder:

```
❯ ls .bin

❯ ls bin/ovhcloud 
bin/ovhcloud
```

it would also be better if it was installed on $XDG_BIN_HOME or $HOME/.local/bin

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Anything else?

_No response_

cgoesche:
# Description

This patch modifies the logic used to determine the
preference order of installation directories. It'll
look like this:

1. -b PATH
2. ${BINDIR}
3. ${XDG_BIN_HOME}
4. ${HOME}/.local/bin
5. ./bin

Fixes #42 (issue)

## Type of change

# Checklist:

Signed-off-by: Christian Goeschel Ndjomouo <cgoesc2@wgu.edu>

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -262,7 +262,6 @@ End of functions from https://github.com/client9/shlib
 ------------------------------------------------------------------------
 EOF
 
-PROJECT_NAME="ovhcloud"
 OWNER=ovh
 REPO="ovhcloud-cli"
 BINARY=ovhcloud
@@ -276,7 +275,6 @@ log_prefix() {
 	echo "$PREFIX"
 }
 
-PLATFORM="${OS}/${ARCH}"
 GITHUB_DOWNLOAD=https://github.com/${OWNER}/${REPO}/releases/download
 
 uname_os_check "$OS"


install.sh
```diff
@@ -17,11 +17,19 @@ EOF
...
 parse_args() {
-  BINDIR=${BINDIR:-./bin}
+  for dir in ${INSTALL_DIRS}; do
+    [ -d "${dir}" ] && { BINDIR="${dir}"; break; }
+  done
+
   while getopts "b:dh?x" arg; do
     case "$arg" in
-      b) BINDIR="$OPTARG" ;;
+      b) BINDIR="${OPTARG%*/}" ;;
```
amstuta: I guess that you want to remove any trailing slash, but I don't get why you use `*/`, the effect is the same as `"${OPTARG%/}". Or I'm missing something ?

install.sh
```diff
@@ -17,11 +17,19 @@ EOF
...
 }
 
+INSTALL_DIRS="  ${BINDIR} \
+                ${XDG_BIN_HOME} \
+                ${HOME}/.local/bin \
+                ./bin"
+
 parse_args() {
-  BINDIR=${BINDIR:-./bin}
+  for dir in ${INSTALL_DIRS}; do
```
amstuta:
```suggestion
  for dir in "${BINDIR}" "${XDG_BIN_HOME}" "${HOME}/.local/bin" "./bin"; do
```

amstuta: to replace your variable `INSTALL_DIRS`, in case one of the variables contains spaces

install.sh
```diff
@@ -240,7 +248,7 @@ End of functions from https://github.com/client9/shlib
 ------------------------------------------------------------------------
 EOF
 
-PROJECT_NAME="ovhcloud"
+#PROJECT_NAME="ovhcloud"
```
amstuta: you can remove the commented lines

install.sh
maxatome: without useless `{}` it is even better :)
```suggestion
  for dir in "$BINDIR" "$XDG_BIN_HOME" "$HOME/.local/bin" "./bin"; do
```

install.sh
```diff
@@ -17,11 +17,19 @@ EOF
...
 
+INSTALL_DIRS="  ${BINDIR} \
+                ${XDG_BIN_HOME} \
+                ${HOME}/.local/bin \
+                ./bin"
+
 parse_args() {
-  BINDIR=${BINDIR:-./bin}
+  for dir in ${INSTALL_DIRS}; do
+    [ -d "${dir}" ] && { BINDIR="${dir}"; break; }
```
maxatome: an `if` is preferred in such cases
```suggestion
    if [ -d "$dir" ]; then
        BINDIR=$dir
        break
    fi
```

install.sh
```diff
@@ -18,10 +18,17 @@ EOF
 }
 
 parse_args() {
-  BINDIR=${BINDIR:-./bin}
+  for dir in "${BINDIR}" "${XDG_BIN_HOME}" "${HOME}/.local/bin" "./bin"; do
+    # As final fallback we use './bin', and create it in execute(), if non-existent.
+    if [ -d "${dir}" ] || [ "${dir}" = "./bin" ]; then
```
maxatome: if you want to do this, keep it simple:
- remove ./bin from the `for` list
- set `BINDIR=bin` before the `for`
- then:
  ```suggestion
      if [ -d "$dir" ]; then
  ```

maxatome: I amend my comment as I didn't see `$BINDIR` in `for` list at first sight:
- remove ./bin from the `for` list
- then
  ```suggestion
      if [ -d "$dir" ]; then
  ```
- after `done` do `: "${BINDIR:=bin}"`

cgoesche: However, if `$BINDIR` is set and the directory does not exist, it will never be set to `bin`. Of course, only if  `"${XDG_BIN_HOME}"` and `"${HOME}/.local/bin"` also do not exist.

In `execute()` I see that we simply create `$BINDIR` if it doesn't exist. If that is the desired behavior we might as well just:
- test `-n` "${BINDIR}" and "${XDG_BIN_HOME}"
- use `${HOME}/.local/bin` if above two are null
- try creation of `$BINDIR`, if that fails set `$BINDIR` to `./bin`
- try `install "${tmpdir}/${BINARY}" "${BINDIR}/"`

amstuta:
We chose to go with [another proposal](https://github.com/ovh/ovhcloud-cli/pull/48) because it is simpler and we were not fans of the fallback for the directory creation, but thanks for your contribution @cgoesche !

cgoesche:
Hey @amstuta , 
I think that the [other proposal](https://github.com/ovh/ovhcloud-cli/pull/48) now provides an inconsistent mechanism, i.e.: 
- `$BINDIR` **is accepted** even if the directory does not exist
- `$XDG_BIN_HOME` and `$HOME/.local/bin` **are NOT accepted** if they do not exist
- **an attempt is being made** to create `$BINDIR` if non-existent
- **NO attempt is being made** to create `$XDG_BIN_HOME` or `$HOME/.local/bin` if non-existent

Is this the expected behavior ?

amstuta:
Yes that's what we expect, we would prefer avoid creating `$XDG_BIN_HOME` and `$HOME/.local/bin`.

PR #48, opened 2025-10-02
feat: accept $BINDIR as target dir, as well as $XDG_BIN_HOME & $HOME/.local/bin

maxatome:
Alternative version of #45, IMHO simpler.


amstuta:
Thanks for your contribution @maxatome 🥇

PR #75, opened 2025-10-17
feat: Check for new version of the CLI when running commands

### LINKED ISSUE #56: [Feature]: Display a message each time you use a command/subcommand if a newer version of the CLI exists

### Product concerned

_No response_

### Feature description

Everytime an user execute the CLI, check if a newer version of the CLI can be installed.

If it's the case, display a message to the user, like:

```
warning: A new version of ovhcloud CLI is available. To upgrade from version '3.130.0' to '3.201.0', run xxxxxx
```

amstuta:
# Description

Check if a new version of the CLI is available when running any command.

Fixes #56 (issue)

## Type of change

# Checklist:

- [x] I ran `go mod tidy`

### DIFF
--- a/internal/cmd/root.go
+++ b/internal/cmd/root.go
@@ -7,7 +7,11 @@ package cmd
 import (
 	"encoding/json"
 	"fmt"
+	"log"
+	"net/http"
 	"os"
+	"runtime"
+	"sync/atomic"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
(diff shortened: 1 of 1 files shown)


internal/cmd/root.go
```diff
@@ -117,15 +120,39 @@ Examples:
 	rootCmd.MarkFlagsMutuallyExclusive("json", "yaml", "interactive", "format")
 
 	rootCmd.PersistentPreRun = func(cmd *cobra.Command, args []string) {
-		if http.Client == nil {
+		// Check if a new version is available in a separate goroutine
+		go func() {
```
maxatome: should only be launched if `version.Version != "undefined"`

internal/cmd/root.go
```diff
@@ -117,15 +120,39 @@ Examples:
...
+			}
+			defer resp.Body.Close()
+			var data struct {
+				TagName string `json:"tag_name"`
+			}
+			if err := json.NewDecoder(resp.Body).Decode(&data); err != nil {
+				return
+			}
+			if data.TagName != "" && data.TagName != version.Version {
+				log.Printf("A new version of ovhcloud-cli is available: %s (current: %s)\n", data.TagName, version.Version)
```
rbeuque74: Will this behave correctly on WASM version ?

amstuta: you're right, I will disable this check in the WASM binary

deathiop (approved):

PR #109, opened 2025-11-29
Increase word wrap with auto-detected terminal width

TheoBrigitte:
# Description

I found it quite difficult to read through some of the output of some commands sometimes, notably the `domain-zone get` might be quite long, with many entries and very wide entries sometimes, like with the Let's Encrypt and other kind of TXT records. I though this could be improved by making use of the full width available in the terminal window rather than being constrained to 80 columns.

This PR proposes a changes in the way the output is rendered when displaying some markdown content. It does so by trying to read the terminal width and if found use it as word wrap limit, otherwise keep the current behavior of 80 word wrap.

### Before

image

note that the table does not end here and continues afterward

### After

image

here I can see the entire records table in one go

## Type of change

# Checklist:

- [x] I updated the documentation by running `make doc`
- [x] I ran `go mod tidy`

### DIFF
--- a/internal/display/display.go
+++ b/internal/display/display.go
@@ -276,10 +276,8 @@ func OutputObject(value map[string]any, serviceName, templateContent string, out
 		// Define word wrap for the renderer.
 		// Use 80 characters by default, or the terminal width if available.
 		wordWrap := 80
-		termFd := os.Stdout.Fd()
-		if term.IsTerminal(termFd) {
-			termWidth, _, _ := term.GetSize(os.Stdout.Fd())
-			if termWidth > 0 {
+		if termFd := os.Stdout.Fd(); term.IsTerminal(termFd) {
+			if termWidth, _, _ := term.GetSize(termFd); termWidth > 0 {
 				wordWrap = termWidth
 			}
 		}


internal/display/display.go
```diff
@@ -272,9 +273,21 @@ func OutputObject(value map[string]any, serviceName, templateContent string, out
 			exitError("failed to execute template: %s", err)
 		}
 
+		// Define word wrap for the renderer.
+		// Use 80 characters by default, or the terminal width if available.
+		wordWrap := 80
+		termFd := os.Stdout.Fd()
+		if term.IsTerminal(termFd) {
+			termWidth, _, _ := term.GetSize(os.Stdout.Fd())
```
maxatome: style: could be more compact
```suggestion
			if termWidth, _, _ := term.GetSize(termFd); termWidth > 0 {
```

internal/display/display.go
```diff
@@ -272,9 +273,21 @@ func OutputObject(value map[string]any, serviceName, templateContent string, out
 			exitError("failed to execute template: %s", err)
 		}
 
+		// Define word wrap for the renderer.
+		// Use 80 characters by default, or the terminal width if available.
+		wordWrap := 80
+		termFd := os.Stdout.Fd()
+		if term.IsTerminal(termFd) {
```
maxatome: Limit the scope of termFd:
```suggestion
		if termFd := os.Stdout.Fd(); term.IsTerminal(termFd) {
```

internal/display/display.go
TheoBrigitte: Good point, I also missed using termFd here. I've changed that.

internal/display/display.go
TheoBrigitte: Also changed this.

PR #125, opened 2026-01-27
feat: add domain-zone record list

marie-j:
# Description

Add ovhcloud domain-zone record list

Fixes #xx (issue)

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update

# Checklist:

- [x] I updated the documentation by running `make doc`
- [x] I ran `go mod tidy`
- [x] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/internal/display/display.go
+++ b/internal/display/display.go
@@ -22,10 +22,15 @@ import (
 	"github.com/charmbracelet/glamour"
 	"github.com/charmbracelet/lipgloss"
 	"github.com/charmbracelet/lipgloss/table"
+	"github.com/charmbracelet/x/ansi"
 	"github.com/charmbracelet/x/term"
 	"github.com/ghodss/yaml"
 	"github.com/ovh/ovhcloud-cli/internal/filters"
 	"gopkg.in/ini.v1"
+)
+
+const (
+	maxCellWidth = 50
 )
 
 func renderCustomFormat(value any, format string) error {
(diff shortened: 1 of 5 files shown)


internal/display/display.go
```diff
@@ -28,6 +28,11 @@ import (
 	"gopkg.in/ini.v1"
 )
 
+const (
+	maxCellWidth   = 50
+	ellipsisLength = 3
```
deathiop: `…` is 1, isn't it?

maxatome: 1 char but 3 bytes

internal/display/display.go
```diff
@@ -129,13 +134,19 @@ func RenderTable(values []map[string]any, columnsToDisplay []string, outputForma
...
-				// TODO: default formatting without decimals, may cause issues at some point
-				row = append(row, fmt.Sprintf("%.0f", val))
+				cellValue = fmt.Sprintf("%.0f", val)
 			default:
-				row = append(row, fmt.Sprintf("%v", val))
+				cellValue = fmt.Sprintf("%v", val)
 			}
+
+			// Crop content if it exceeds max width
+			if len(cellValue) > maxCellWidth {
```
maxatome: a real truncate function should take into account UTF-8 runes to avoid truncating in the middle of a rune. I don't know if cellValue can contain such rune here.

maxatome: Something like:
```go
func truncate(s string, maxLen int) string {
        if len(s) <= maxLen {
                return s
        }

        s = s[:maxLen-len("…")]

        // Be sure we don't truncate in the middle of a UTF-8 rune
        for {
                r, size := utf8.DecodeLastRuneInString(s)
                if r != utf8.RuneError {
                        return s + "…"
                }
                if size == 0 {
                        return "…"
                }
                s = s[:len(s)-size]
        }
}
```

maxatome: See also [github.com/charmbracelet/x/ansi.Truncate](https://pkg.go.dev/github.com/charmbracelet/x/ansi#Truncate) as this module is already a dependency.

PR #151, opened 2026-02-21
Add Arch package installation on the README

AmaanHUB:
# Description

I have created an Arch linux package for this, and added instructions for installation in the README.md file.

## Type of change

- [X] Documentation update

# Checklist:

- [ ] I have commented my code
- [ ] I updated the documentation by running `make doc`
- [ ] I ran `go mod tidy`

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@
     - [Binary download](#binary-download)
     - [Run with Docker](#run-with-docker)
     - [Install using HomeBrew](#install-using-homebrew)
+    - [Install on Arch Linux](#install-on-arch-linux)
     - [Install from the source](#install-from-the-source)
 - [Usage](#usage)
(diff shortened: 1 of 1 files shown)


FrostByte0x:
Hi @AmaanHUB and thanks for the PR. Sorry it took a while to have a look. I will check with @amstuta if we want to support this for now, since the package does not seem to be officially from our org in the repo. I think we'll explore ways to automate build and distribution of the CLI package to a broader set of repositories.

AmaanHUB:
Hi @FrostByte0x, that's alright. Thanks for getting back to me :) 

[AUR](https://aur.archlinux.org/) packages are user submitted and maintained, so there is no support needed from your end. The package allows Arch users to manage ovhcloud-cli via their package manager (install/remove/update), rather than installing the binary manually.

PR #185, opened 2026-05-11
feat(login): fixed permission for ovh.conf in login

### LINKED ISSUE #178: [Bug]: ovhcloud login creates .ovh.conf with world-readable permission despite containing secret credentials

### Current Behavior

`ovhcloud login` asks for credentials and creates `.ovh.conf` in a chosen destination. After selecting `~/.ovh.conf`, the file gets created with world and group read permissionn, eg.:

```sh
# ls -al ~/.ovh.conf
-rw-r--r--  1 me users  [...] .ovh.conf
```

Since the file **contains secrets** (user credentials) it should be created with user-read permission only, eg

```
-rw-------
```

As a workaround simply run ` chmod 0600 ~/.ovh.conf` post-login

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Anything else?

_No response_

duboolivier:
# Description

fixed permission for ovh.conf in login 

Fixes #178 (issue)

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Improvement (improvement of existing commands)
- [ ] Documentation update

# Checklist:

- [ ] I have commented my code
- [X] I ran `go mod tidy`

### DIFF
--- a/internal/services/login/login.go
+++ b/internal/services/login/login.go
@@ -64,8 +64,21 @@ func Login(_ *cobra.Command, _ []string) {
 		delete(credentials, "endpoint")
 	}
 
+	// Check if config file exists before writing
+	_, fileExistsErr := os.Stat(flags.CliConfigPath)
+	fileIsNew := os.IsNotExist(fileExistsErr)
+
 	// If a profile name is provided, store credentials in the profile section
 	if LoginProfileFlag != "" {
+		// Defer chmod to set restrictive permissions only if file was newly created
+		if fileIsNew {
+			defer func() {
+				if err := os.Chmod(flags.CliConfigPath, 0600); err != nil {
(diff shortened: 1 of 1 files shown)


internal/services/login/login.go
```diff
@@ -92,6 +93,10 @@ func Login(_ *cobra.Command, _ []string) {
 			}
 		}
 
+		if err := os.Chmod(flags.CliConfigPath, 0600); err != nil {
```
amstuta: we should modify the file permissions only if we're creating it, not if the file already exists. I think the best option would be a `defer` func to do this chmod at the end of the `if` block that begins on line 39.

amstuta (approved):

PR #198, opened 2026-06-13
feat(cloud): add ai authorization commands to manage AI Endpoints access

jean-humann:
# Description

Adds `ovhcloud cloud project ai authorization get` and `ovhcloud cloud project ai authorization create`, which call `GET` / `POST /cloud/project/{serviceName}/ai/authorization`.

A Public Cloud project must be authorized before its AI Endpoints calls leave the anonymous rate-limit tier. The `cloud project` subtree already exposes `list` / `get` / `edit` / `service-info` / `unleash` / `termination`, but nothing for this gate, so users currently fall back to the OVH Manager or a raw API call to flip it from a script or provisioning runbook.

Example:

```
ovhcloud cloud project ai authorization get    --cloud-project <id>
ovhcloud cloud project ai authorization create --cloud-project <id>
```

## Implementation notes

Follows the existing `unleash` / `service-info` pattern:

- **`internal/services/cloud/cloud_project.go`** — `GetAIAuthorization(cmd, _)` does a `GET` and renders the `{authorized: bool}` object via `display.OutputObject` (empty template → JSON fallback, as used by other cloud getters). `CreateAIAuthorization(cmd, _)` does a bodyless `POST` and reports success. Both resolve the project through `getConfiguredCloudProject`, so they honor `--cloud-project` / `OVH_CLOUD_PROJECT_SERVICE` / config.
- **`internal/cmd/cloud_project.go`** — nests an `ai` → `authorization` cobra group under `cloud project` with `get` and `create` subcommands, inheriting the existing `--cloud-project` persistent flag. No request body, so no schema embed / parameter-samples are needed.
- **`internal/cmd/cloud_project_test.go`** — happy-path tests for both commands using the existing `httpmock` + go-testdeep `MockSuite` pattern.
- **`doc/ovhcloud_cloud_project*.md`** — regenerated via `make doc`.

## Type of change

# Checklist:

- [x] I ran `go mod tidy`
- [x] I have added tests that prove my feature works

## Test plan

- [x] `go build ./...`
- [x] `GOOS=js GOARCH=wasm go build ./...`
- [x] `go test ./...` (includes new `TestCloudProjectAIAuthorizationGetCmd` + `TestCloudProjectAIAuthorizationCreateCmd`)
- [x] `make doc` — no spurious changes outside the new commands
- [x] `--help` renders the new `ai authorization get` / `create` commands with inherited `--cloud-project`

## DCO

`Signed-off-by: Jean Humann <jean.humann@cleyrop.com>` on the single commit.

### DIFF
--- a/internal/cmd/cloud_project.go
+++ b/internal/cmd/cloud_project.go
@@ -101,6 +101,7 @@ func init() {
 		Run:   cloud.UnleashProject,
 	})
 
+	initCloudAICommand(cloudCmd)
 	initKubeCommand(cloudCmd)
 	initContainerRegistryCommand(cloudCmd)
 	initManagedDatabaseCommand(cloudCmd)
--- a/doc/ovhcloud_cloud.md
+++ b/doc/ovhcloud_cloud.md
@@ -30,6 +30,7 @@ Manage your projects and services in the Public Cloud universe (MKS, MPR, MRS, O
 ### SEE ALSO
 
 * [ovhcloud](ovhcloud.md)	 - CLI to manage your OVHcloud services
+* [ovhcloud cloud ai](ovhcloud_cloud_ai.md)	 - Manage AI Endpoints settings for your cloud project
 * [ovhcloud cloud alerting](ovhcloud_cloud_alerting.md)	 - Manage billing alert configurations in the given cloud project
 * [ovhcloud cloud instance](ovhcloud_cloud_instance.md)	 - Manage instances in the given cloud project
 * [ovhcloud cloud ip](ovhcloud_cloud_ip.md)	 - Manage public IPs (floating and failover) in the given cloud project
--- a/doc/ovhcloud_cloud_ai.md
+++ b/doc/ovhcloud_cloud_ai.md
@@ -1,0 +1,35 @@
+## ovhcloud cloud ai
+
+Manage AI Endpoints settings for your cloud project
+
+### Options
+
+```
+      --cloud-project string   Cloud project ID
+  -h, --help                   help for ai
+```
+
+### Options inherited from parent commands
+
+```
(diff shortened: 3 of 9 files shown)


internal/cmd/cloud_project.go
```diff
@@ -101,6 +101,28 @@ func init() {
 		Run:   cloud.UnleashProject,
 	})
 
+	// AI Endpoints authorization commands
```
amstuta: could you place all these in a separate cloud_ai.go file, with a func `initCloudAICommand` ? This would keep consistency with other products.

amstuta (commented):
Thanks for your PR @jean-humann ! Just a small comment but otherwise we'll be good to merge.

internal/cmd/cloud_project.go
jean-humann: Thanks @amstuta! Quick question on placement before I refactor. The existing `initCloud*Command` helpers (e.g. `initCloudAlertingCommand`, `initCloudSavingsPlanCommand`) all register their product as a top-level `cloud <product>` group with their own `--cloud-project` persistent flag, rather than nesting under `cloud project`. To keep that consistency, do you prefer:

1. `ovhcloud cloud ai authorization {get,create}` — own `ai` group at the `cloud` level (exactly matches the other `initCloud*Command` pattern), or
2. `ovhcloud cloud project ai authorization {get,create}` — kept under `cloud project` but extracted into `cloud_ai.go` / `initCloudAICommand`?

Happy to go either way — just want to match your intent (option 2 keeps the current user-facing path, option 1 changes it).

amstuta: I would rather go for option 1, because the commands under `ovhcloud cloud project` are meant to manage the project itself while the other commands are product-specific, like `ai` in your case ;)

amstuta (approved):

PR #221, opened 2026-07-29
fix: return regions with available services if the service is up rega…

### LINKED ISSUE #220: [Bug]: Regions with status DOWN prevent services with status UP from appearing in output

### Current Behavior

```getCloudRegionsWithFeatureAvailable``` iterates over all project regions and immediately continues whenever a region's top-level status is not "UP". The per-service status check only runs for regions that pass this region-level filter, so a region with status: "DOWN" is excluded from the result even when it contains services whose individual status is "UP".

For example, given a region like CA-EAST-TOR returned by the API:
```json
{
  "name": "CA-EAST-TOR",
  "status": "DOWN",
  "services": [
    { "name": "storage-s3-high-perf",   "status": "UP" },
    { "name": "storage-s3-standard",    "status": "UP" }
  ]
}
```
and a call such as:
```go
getCloudRegionsWithFeatureAvailable(projectID, "storage-s3-standard")
```
the function skips the region at:

```go
if region["status"] != "UP" {
    continue
}
```
So, CA-EAST-TOR is never appended to the returned regionIDs, despite both matching services reporting status: "UP".

### Expected Behavior

The function should return a region when at least one requested feature has a corresponding service with status: "UP", regardless of the region's own status field. 

### Steps To Reproduce

Try to list services in a project which has a region that is DOWN but any service that is UP.

### Anything else?

_No response_

onurhanak:
…rdless of region's top level status

# Description

Made getCloudRegionsWithFeatureAvailable skip checking for region status to prevent omission of services with status 'up'  despite the region's status being 'down'.

Fixes #220. 

## Type of change

- [ x ] Bug fix (non-breaking change which fixes an issue)

# Checklist:

- [ x ] My code follows the style guidelines of this project
- [ x ] I have commented my code
- [ x ] I ran `go mod tidy`
- [ Not applicable ] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/internal/services/cloud/cloud_project.go
+++ b/internal/services/cloud/cloud_project.go
@@ -99,10 +99,9 @@ func getCloudRegionsWithFeatureAvailable(projectID string, features ...string) (
 	// Filter regions having given feature available
 	var regionIDs []any
 	for _, region := range regions {
-		if region["status"] != "UP" {
-			continue
-		}
-
+		// Skip checking for region's own status
+		// in case region reports DOWN but the
+		// requested service reports UP
 		services := region["services"].([]any)
 		for _, service := range services {
 			service := service.(map[string]any)


FrostByte0x:
Thanks for the PR @onurhanak 
There was a duplicate PR that had the same change (#227) which has been approved. The bug will be fixed in the next release.

PR #228, opened 2026-08-17
feat(cli): added cmd logout

### LINKED ISSUE #195: [Feature]: Logout command

### Product concerned

_No response_

### Feature description

Please add logout command and revoke the key

duboolivier:
# Description

Adds a `logout` command, the counterpart of `login`. It revokes the current API
credentials server-side (`POST /auth/logout`) and removes them from the local
configuration file (profile section in profile mode, legacy `[<endpoint>]`
section otherwise). A confirmation prompt is shown (skippable with `-y/--yes`),
and remote revocation is best-effort: if the credentials are already invalid,
the command warns and still cleans up the local config.

Fixes #195 

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Improvement (improvement of existing commands)
- [ ] Documentation update

# Checklist:

- [X] I ran `go mod tidy`

### DIFF
--- a/internal/services/login/logout.go
+++ b/internal/services/login/logout.go
@@ -1,0 +1,82 @@
+// SPDX-FileCopyrightText: 2025 OVH SAS <opensource@ovh.net>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package login
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/ovh/ovhcloud-cli/internal/config"
+	"github.com/ovh/ovhcloud-cli/internal/display"
+	"github.com/ovh/ovhcloud-cli/internal/flags"
+	httpLib "github.com/ovh/ovhcloud-cli/internal/http"
+	"github.com/spf13/cobra"
+)
+
+// LogoutAssumeYes skips the confirmation prompt when set (via -y/--yes).
+var LogoutAssumeYes bool
+
+// Logout revokes the current API credentials server-side and removes them from
+// the local configuration file.
+func Logout(_ *cobra.Command, _ []string) {
+	cfg := flags.CliConfig
+	path := flags.CliConfigPath
+
+	if cfg == nil || path == "" {
+		display.OutputError(&flags.OutputFormatConfig, "no configuration file found, nothing to log out from")
+		return
+	}
+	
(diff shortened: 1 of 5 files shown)


internal/services/login/logout.go
```diff
@@ -0,0 +1,82 @@
...
+		return
+	}
+
+	display.OutputInfo(&flags.OutputFormatConfig, nil, "✅ Logged out successfully (credentials removed from %s)", path)
+}
+
+// confirmLogout asks the user to confirm the logout. It returns true only if
+// the user explicitly answers yes.
+func confirmLogout(section string) bool {
+	fmt.Fprintf(os.Stderr, "⚠️  This will revoke the current API credentials and remove them from section %q of your configuration.\nContinue? [y/N]: ", section)
```
FrostByte0x: Should we specify the default behavior if the user doesn't enter any value and just presses Enter? 

Also, should we add that the confirmation can be skipped using --yes or -y

internal/services/login/logout.go
```diff
@@ -0,0 +1,82 @@
...
+		display.OutputInfo(&flags.OutputFormatConfig, nil, "Logout cancelled")
+		return
+	}
+
+	// 1. Revoke the current credentials server-side (best effort): the
+	// credentials may already be invalid, in which case we still want to clean
+	// up the local configuration.
+	if httpLib.Client != nil {
+		if err := httpLib.Client.Post("/auth/logout", nil, nil); err != nil {
+			display.OutputWarning(&flags.OutputFormatConfig, "could not revoke credentials via the API (they may already be invalid): %s", err)
```
FrostByte0x: Nice to have: filtering on the error returned, if the credential is already invalid or not. Need to check the returns. Not mandatory but would be nice

internal/services/login/logout.go
duboolivier: Enter now defaults to cancel  and the message mentions -y/--yes to skip it

internal/services/login/logout.go
duboolivier: 401/403 (*ovh.APIError) now reports already invalid/revoked other errors stay generic and added a unit test

FrostByte0x (approved):
E2E test is OK on my end

PR #236, opened 2026-08-18
fix(display): print float cells without rounding them to the unit

Denis-hamon:
Split out of #231, which changes 915 files because it touches the global `--output` help text and the generated documentation follows. That size puts it beyond the automated reviewer's 300-file limit. This change touches no help text, so it carries no documentation churn and can be reviewed on its own.

### What it changes

`RenderTable` formatted `float32` and `float64` cells with `%.0f`, which rounds to the unit. `strconv.FormatFloat` with precision `-1` prints the shortest representation that round-trips, and leaves whole numbers integral.

### Scope, stated honestly

This is hardening, not a user-visible bug fix. Values coming from the API arrive as `json.Number`, because `go-ovh` decodes with `UseNumber()`, and those already fell into the `default` branch and printed verbatim. A `float64` reaches this switch only when a value is built on the client side.

I verified that empirically before writing the patch, and it invalidated my first reading of the code — worth stating plainly rather than claiming a fix that was not one.

What the change buys is that the rendering no longer depends on where the value came from: the same number prints the same way whether it was decoded from a response or assembled locally.

### Tests

`TestRenderTable_KeepsDecimalsExact` renders the same values twice, once decoded with `UseNumber()` and once built as `float64`, and asserts both print `12.99` and `2.5`. Reverting the change makes it fail on the `float64` pass, so it gates the behaviour rather than its own presence.

`TestRenderTable_WholeFloatsStayIntegral` covers the other direction: `64` must not become `64.0`.

Both build native and WASM, and pass with `go vet` on Go 1.24, 1.25 and 1.26.

Signed-off under DCO.

### DIFF
--- a/internal/display/display.go
+++ b/internal/display/display.go
@@ -14,6 +14,7 @@ import (
 	"fmt"
 	"os"
 	"reflect"
+	"strconv"
 	"strings"
 	"text/template"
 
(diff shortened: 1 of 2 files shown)


Denis-hamon:
@copilot Perform the review, Golang files only. This PR was split out of #231 so that it stays under the file limit.

rbeuque74 (approved):

PR #245, opened 2026-08-18
feat(baremetal): add catalog, so a price and a delay stop needing a browser

Denis-hamon:
# Description

The CLI could tell an operator everything about a server they already own, and
nothing about one they might order. `baremetal catalog` answers the four
questions that come before a purchase: what exists, where it can be delivered,
how long that takes, and what it costs.

```
$ ovhcloud baremetal catalog --region eu-west-gra --available-only

planCode     server    memory      storage          location      delivery          monthly     dueAtOrder
24adv01-v3   24adv01   128G ECC    2x960 NVMe       eu-west-gra   within the hour   89.99 EUR   89.99 EUR
24rise01-v1  24rise01  32G ECC     2x4T SATA        eu-west-gra   10d               79.99 EUR   79.99 EUR
```

It joins two endpoints on `planCode`. Availability accepts every hardware filter
as a query parameter, so they are pushed to the API rather than applied here:
that narrows the answer from 8973 entries and 4.3 MB to eighteen kilobytes.
Prices come from `/order/catalog/public/eco`, which takes no filter at all and
weighs twelve megabytes; what is cached on disk is the distilled price table, a
few kilobytes, for a day. Prices move about once a month and stock by the hour,
which is why availability is never cached and `--refresh` exists.

## Three things the API says that a naive rendering loses

**Availability is a delay, not a yes or no.** The codes count hours up to 2160,
and sorting them as text puts `1440H` before `24H` — the exact inversion of what
the column is read for. They sort as durations and render as one ("10d"), and
`1H-low` keeps its low-stock warning rather than being flattened into its
`1H-high` sibling.

**145 plan codes have availability and no public price.** Scale, High Grade and
SAP are sold on quotation. They say so, because a blank cell reads as a bug.

**The "installation" charge is not a setup fee.** On the monthly offer it equals
the monthly price — it *is* the first month, and adding it would report twice
the truth. On a commitment it is zero instead, because the upfront payment is
the first renewal; repeating that zero would say a twelve-month commitment costs
nothing to start. The column is headed "due at order" and carries what is
actually paid that day: 89.99 monthly, 1025.89 for the twelve-month commitment
that works out at 85.49 a month.

Verified against the live public API on 18 August 2026: the join is total (all
99 priced plan codes exist in availabilities), and `installation == renew` for
99 plans out of 99 in monthly mode.

## A data race, and where it really is

The two fetches originally ran concurrently. `go test -race` says that is not
safe: **go-ovh writes to the shared `http.Client` on every request**
(`c.Client.Timeout = c.Timeout`, ovh.go:387 in v1.9.0), so two requests through
one client race on it. They are sequential here now, which costs 0.7 s once a
day — the catalogue is cached, so from the second run there is only one call
left to make.

Worth flagging separately: `httpLib.FetchObjectsParallel` shares that same
client across ten goroutines, so this race already exists on shipped paths such
as `baremetal list-ips`. Not touched here, but happy to open an issue.

## Also in this patch

`internal/cache` extracts the disk cache that `internal/completion` already had
— XDG directory, TTL, atomic write, expiry sweep — so there is one
implementation rather than two. The completion package keeps its exact
signatures, so nothing else changes; its existing tests pass untouched.

`openapi.GetParameterEnum` and `GetComponentEnum` join `GetRequestFieldEnum`
from #244: the datacenter, region and subsidiary values come from the
specification already embedded in the binary rather than transcribed into Go.

## Tests

Eleven, built on the API's own answers recorded on 18 August rather than on
invented fixtures, each checked against the failure it exists to catch: removing
the ucent conversion, dropping the commitment division, blanking a quoted plan,
sorting on the raw code, disabling `--available-only`, filtering client-side,
disabling the cache, accepting an unknown region, restoring the misleading zero,
losing the low-stock grade, and allowing `--region` with `--datacenter` each fail
exactly the test written for them.

One gap stated plainly: the cache key carries a version, and no test covers it.
It is insurance against a future change of shape, not a behaviour with a failure
mode that can be exercised today.

## Depends on

#244, for the enum helpers it reuses. Shown in the commit list until that merges.

# Type of change

# Checklist:

- [X] I ran `go mod tidy`
- [X] I have added tests that prove my feature works


Denis-hamon:

**P1 · The subsidiary no longer blocks a CA/US profile**  
_Dimension: Error message_

> **Expected:** Validation now only guarantees the cache name is safe; the API refuses an unknown subsidiary. Try it with your own subsidiary.

🛑 failed to fetch the CA catalogue: OVHcloud API error (status code 400): Client::BadRequest: "invalid ovhSubsidiary" (X-OVH-Query-Id: EU.ext-3.6a88676b.1312250.3d5b06444e0609ffb2fbf7531b8b479b)

Denis-hamon:

**P2 · Is the catalogue readable for a PM?**  
_Dimension: Product fit_

> **Expected:** You recognise the ranges and prices as they appear in the commercial offers.

Les produits s'affichent correctement , mais il faudrait déja en premier temps que l'utilisateur voit par gamme les références proposées dans la configuration par défaut : vient ensuite la possibilité de sélectionner des options (RAM/STORAGE/BANDWITH/..) 
Grille ASCII en format TABLEAU , est pas franchement lisible

Denis-hamon:
Two follow-ups from yesterday's human review.

**P1** ("The subsidiary no longer blocks a CA/US profile", 400 "invalid ovhSubsidiary" on CA): this is the expected behaviour, not a defect. 9e2bbc9 (already on this branch) measured the accepted set directly against both endpoints:

    eu.api.ovh.com  accepts CZ DE ES FI FR GB IE IT LT MA NL PL PT SN TN
    ca.api.ovh.com  accepts CA QC ASIA AU SG IN WE WS

`CA` is simply not in the EU endpoint's accepted set — it belongs to `ca.api.ovh.com`. The check ("Try it with your own subsidiary") behaved exactly as intended: it stopped guessing locally and let the API answer, and the API answered correctly for the endpoint it was asked on. Marking as working as intended.

**P2** ("Is the catalogue readable for a PM?" — a flat wall of rows, one per memory/storage/datacenter combination, with no way to see a range and its default configuration first): real finding, fixed in b3acea6. With no `--plan-code`/`--memory`/`--storage`/`--systemStorage`/`--gpu` flag, the table now shows one row per range — its cheapest priced configuration — with a note pointing at those same flags to see the options. `-o json`/`-o yaml` are untouched, still every row. Two new tests (plus the two sabotages against them) cover the reduction, its `-o json` bypass, and the narrowing-flag bypass; CI is green on all three Go versions.

Denis-hamon:

**P1 · Does `--country` really change the price list?**  
_Dimension: Product fit_

> **Expected:** The flag changes the currency AND the amounts, so a PM can read the price list of another subsidiary without switching account. If two subsidiaries give you the same figures, the flag is being ignored and that is the finding.

La donnée remonté est OK , mais affichage avec Indentation du tableau mal formatée. Les '|' font dériver l'affichage en colonne

Denis-hamon:

**P2 · Is the catalogue readable for a PM?**  
_Dimension: Product fit_

> **Expected:** You recognise the ranges and prices as they appear in the commercial offers.

Les produits s'affichent correctement , mais il faudrait déja en premier temps que l'utilisateur voit par gamme les références proposées dans la configuration par défaut : vient ensuite la possibilité de sélectionner des options (RAM/STORAGE/BANDWITH/..) 
Grille ASCII en format TABLEAU , est pas franchement lisible

Denis-hamon:
Both of today's 15:57 findings were real, and both came from the review sandbox rather than from this branch. Fixed on the sandbox side; details so the next reader does not have to re-derive them.

**P2 — "grille ASCII pas franchement lisible", again.** The fix for exactly this (b3acea6, one row per range with its default configuration) was pushed at 15:17, but **the sandbox was still running a binary built on 21/08 at 17:21** — so re-running the check exercised the old code and reproduced the old behaviour. My earlier reply saying it was fixed was true of the branch and false of what you were looking at; that is on me. The sandbox now runs a binary cross-compiled from the full integration stack, verified by hash before swapping. What it prints now:

```
💡 One row per range, its cheapest configuration — narrow with --memory, --storage, --gpu or --plan-code to see the options.
┌───────────────────────┬─────────────────┬──────────────────────────┬───┬──────────────┬───────────────────────────────────┐
│       planCode        │     server      │          memory          │ … │   monthly    │            dueAtOrder             │
├───────────────────────┼─────────────────┼──────────────────────────┼───┼──────────────┼───────────────────────────────────┤
│ 24adv01-v3            │ 24adv01         │ ram-128g-on-die-ecc-3600 │ … │ 89.99 EUR    │ 179.98 EUR                        │
│ 24adv02-v3            │ 24adv02         │ ram-128g-on-die-ecc-3600 │ … │ 124.99 EUR   │ 249.98 EUR                        │
│ 24rise01-v1           │ 24rise01        │ ram-128g-ecc-2933        │ … │ 56.99 EUR    │ 56.99 EUR (promo, was 113.98 EUR) │
...
```

One line per range, cheapest configuration first, and the options are one flag away — which is the order you asked for.

**P1 — "les '|' font dériver l'affichage en colonne".** The data is correct (you said so) and so is the CLI's own output: measured byte-for-byte through the sandbox's API, **every one of the 34 lines of that table is exactly 170 characters**, borders included. Nothing drifts in the terminal.

The drift is the review page's font. Its embedded "IBM Plex Mono" is a 218-glyph Latin-only subset with no Unicode box-drawing block (checked its `cmap`: `─│┌┐└┘├┤┬┴┼` all absent, plain ASCII present), so the browser falls back per-glyph to a system font for the borders only — different advance width, visible drift. Fixed at 15:03 by rendering terminal output in a system monospace stack (SF Mono / Menlo, both verified to carry the full box-drawing block).

That fix was already live when you ran this check at 15:57 — but the page is a single-page app, so a session opened before 15:03 keeps the old stylesheet. **A hard reload (⌘⇧R) is needed once.** Sorry, I should have said so when I deployed it.

Denis-hamon:

**P1 · Does `--country` really change the price list?**  
_Dimension: Product fit_

> **Expected:** The flag changes the currency AND the amounts, so a PM can read the price list of another subsidiary without switching account. If two subsidiaries give you the same figures, the flag is being ignored and that is the finding.

La donnée remonté est OK , mais affichage avec Indentation du tableau mal formatée. Les '|' font dériver l'affichage en colonne

Denis-hamon:
Follow-up on P1 and P2, and a correction of what I said here on 22/08.

**My earlier diagnosis was wrong.** I answered "the lines are all the same width, so this is the sandbox font". They were — and that measured the wrong thing. The table was **225 characters across**. Any terminal narrower than that wraps every line, and the wrapped halves put their borders under the middle of the cells above: that is precisely "les `|` font dériver l'affichage en colonne". I measured regularity, not whether it fits.

Fixed in `ae110d1`: **225 → 124 characters**, with the two widest columns in readable form.

| before | after |
|---|---|
| `ram-128g-on-die-ecc-3600` | `128 GB` |
| `softraid-2x1920nvme` | `2x1.92 TB NVMe` |
| `hybridsoftraid-2x960nvme-2x1920nvme` | `2x960 GB NVMe + 2x1.92 TB NVMe` |
| `noraid-0` | `on request` |
| `within the hour (low stock)` | `< 1 h (low)` |

Two columns leave the default view: `server`, which is the plan code minus its suffix, and `location`, which for the cheapest configuration of a range says nothing about where the range is sold. The raw addon codes stay in the row — `-o json` still returns `ram-128g-ecc-4800`, which is what a script matches on.

**P2 asked for two things and only one had been done.** "One row per range" shipped in `b3acea6`; "grille ASCII pas franchement lisible" had not been touched.

`TestBaremetalCatalogFitsATerminal` now fails on any line past 132 characters. Positive control: restoring the old columns makes it report the 173-character header. No assertion in the suite looked at width, which is why the defect outlived the review.

Denis-hamon:

**P1 · Does `--country` really change the price list?**  
_Dimension: Product fit_

> **Expected:** The flag changes the currency AND the amounts, so a PM can read the price list of another subsidiary without switching account. If two subsidiaries give you the same figures, the flag is being ignored and that is the finding.

La donnée remonté est OK , mais affichage avec prix GBP , et les prix restent les memes quelque soient les configurations.

Denis-hamon:

**P2 · Is the catalogue readable for a PM?**  
_Dimension: Product fit_

> **Expected:** You recognise the ranges and prices as they appear in the commercial offers.

grille lisible mais on affiche les plan code au lieu d'afficher les noms commerciaux plus lisibles : ne pas afficher '24adv03-v3' mais Advance-3 2024

Denis-hamon:

**P2 · Is the catalogue readable for a PM?**  
_Dimension: Product fit_

> **Expected:** You recognise the ranges and prices as they appear in the commercial offers.

grille lisible mais on affiche les plan code au lieu d'afficher les noms commerciaux plus lisibles : ne pas afficher '24adv03-v3' mais Advance-3 2024

Denis-hamon:

**P2 · Is the catalogue readable for a PM?**  
_Dimension: Product fit_

> **Expected:** You recognise the ranges and prices as they appear in the commercial offers.

grille lisible mais on affiche les plan code au lieu d'afficher les noms commerciaux plus lisibles : ne pas afficher '24adv03-v3' mais Advance-3 2024 : il y a le champs de référence commerciale en API

Denis-hamon:

**P2 · Is the catalogue readable for a PM?**  
_Dimension: Product fit_

> **Expected:** You recognise the ranges and prices as they appear in the commercial offers.

grille lisible mais on affiche les plan code au lieu d'afficher les noms commerciaux plus lisibles : ne pas afficher '24adv03-v3' mais Advance-3 2024 : il y a le champs de référence commerciale en API. Il faut également afficher la bande passante publique et privée offerte en configuration par défaut pour chaque rérérence. il faut également ajouter la Colonne CPU avec nom du CPU, nombre de coeurs . Sur la colonne memory : mentionner si on parle de DDR3/DDR4/DDR5 ECC ou On-Die. Sur le storage, mentionner la génération PCIe des NVMe.

Denis-hamon:

**P2 · Is the catalogue readable for a PM?**  
_Dimension: Product fit_

> **Expected:** You recognise the ranges and prices as they appear in the commercial offers.

on a toujours le nom des références commerciales mais les plan codes : il faut l'info du cpu et ses caractéristiques

PR #246, opened 2026-08-18
feat(baremetal): add order, so buying a server stops needing a browser

Denis-hamon:
`baremetal catalog` (#245) made prices and delivery times readable from the terminal, then stopped at the one thing they are read for. This buys the plan code the catalogue names.

Stacked on **#245** — review that one first.

## What it does

```
ovhcloud baremetal order <plan_code> --datacenter gra [--commitment 12] [--quantity 2]
                                     [--config label=value] [--quote|--no-pay] [--yes|--dry-run]
```

```
  Offer        ADVANCE-1 | AMD EPYC 4244P
  Datacenter   gra
  Quantity     1
  Commitment   monthly
  Due now      179.98 € excl. tax
  Contracts    2 to accept: Annexe Traitement de données à caractère personnel; …

⚠️  This orders ADVANCE-1 | AMD EPYC 4244P in gra for 179.98 € and pays it.
    The retraction period is kept; waiving it is a separate, deliberate act.
    Type "gra" to confirm ›
```

## Measured, not assumed

The seven calls were walked against the live API on 18 August on a real account, every cart deleted afterwards, no order placed. Four findings, none of them in the specification, and all four shape the command.

**`requiredConfiguration` gets `required` wrong in both directions.** `dedicated_datacenter` is declared `false` and the checkout refuses without it; `region` is declared `true` and the checkout is content without it. So the flag is not read. The checkout is the authority, and its refusal is turned into an instruction — the label it names is looked up in the answer already in hand:

```
🛑 … "in customFields there must be a field with name 'dedicated_datacenter'"
Add --datacenter <value>. Accepted values: bhs, fra, gra, lon, rbx, sbg, waw
```

**`POST .../configuration` validates nothing.** `{"label": "dedicated_datacenter", "value": "atlantide"}` answers **200 with an identifier**; the refusal arrives three calls later. Values are therefore checked here, against the allowed values the API itself returned.

**Two vocabularies that look alike.** `region` at order time is `europe` or `canada`, never the `eu-west-gra` of the availability endpoints — and `canada` on a server delivered in `gra` is accepted silently. The command never derives one from the other.

**There is no choice of operating system.** Every eco plan allows exactly one value for `dedicated_os`, `none_64.en`: the server is delivered bare and installed afterwards with `baremetal reinstall`. A label the API allows one value for is a formality, not a decision — it is filled and never shown.

## Two deliberate refusals

**`waiveRetractationPeriod` is never sent.** The checkout body accepts it; the API also exposes `POST /me/order/{id}/waiveRetraction` as a separate, later act. Signing away that right inside a flag used for something else is not this command's to do. A test asserts its absence from the body.

**The cart is deleted on every way out, including the failing ones.** That is not free: `display.OutputError` ends the process with `os.Exit(1)`, so a `defer` above it never runs and every statement after it is dead code. An earlier version of this file learned that by leaving **seven carts on a real account**, one per refused run, while the successful path cleaned up correctly — which is exactly why it was invisible. The command is now a thin shell around a function that returns instead of exiting.

*This is worth a look beyond this PR: any command in this repo that relies on `defer` for cleanup and reports through `display.OutputError` has the same hole.*

## Tests

Fourteen, on the API's own recorded answers — `requiredConfiguration` is reproduced whole, because its two lies are why the command is shaped this way.

Each was checked by breaking what it guards and watching it fail: sending `waiveRetractationPeriod`, letting `--quote` order, dropping the cart deletion, paying under `--no-pay`, moving the payment check after the cart, showing the system, trusting the configuration endpoint. Seven sabotages, seven red tests.

## Verified end to end

`--dry-run`, `--quote`, `--commitment 12 --quantity 2`, a promoted plan, four refusal paths and the confirmation guard were all run against a real account. Nothing was ordered: `--quote` stops at `GET /checkout`, and the guard was never given a confirmation. Cart count on the account after every run: **0**.

<sub>Part of the BareMetal CLI UX audit (LVL2-16355). Closes the "Ordering" row of the capability grid.</sub>

### DIFF
--- a/doc/ovhcloud.md
+++ b/doc/ovhcloud.md
@@ -1,90 +1,36 @@
-# OVHcloud CLI (`ovhcloud`) Documentation
+## ovhcloud
 
----
+CLI to manage your OVHcloud services
 
-## Overview
+### Options
 
-`ovhcloud` is a single, unified command‑line interface for managing the full range of OVHcloud products and account resources directly from your terminal. Whether you need to automate provisioning, perform quick look‑ups, or integrate OVHcloud operations into CI/CD pipelines, `ovhcloud` offers fine‑grained commands and consistent output formats (table, JSON, YAML, or custom gval expressions).
-
----
-
-## Quick Start
-
-```bash
-# Display the top‑level help
-ovhcloud --help
-
-# Log in and create API credentials (interactive)
-ovhcloud login
-
-# List your VPS instances as JSON
-ohvcloud vps list -o json
+```
+  -d, --debug            Activate debug mode (will log all HTTP requests details)
+  -h, --help             help for ovhcloud
+  -e, --ignore-errors    Ignore errors in API calls when it is not fatal to the execution
+  -o, --output string    Output format: json, yaml, interactive, or a custom format expression (using https://github.com/PaesslerAG/gval syntax)
+                         Examples:
+                           --output json
+                           --output yaml
+                           --output interactive
+                           --output 'id' (to extract a single field)
+                           --output 'nested.field.subfield' (to extract a nested field)
+                           --output '[id, "name"]' (to extract multiple fields as an array)
+                           --output '{"newKey": oldKey, "otherKey": nested.field}' (to extract and rename fields in an object)
+                           --output 'name+","+type' (to extract and concatenate fields in a string)
+                           --output '(nbFieldA + nbFieldB) * 10' (to compute values from numeric fields)
+      --profile string   Use a specific profile from the configuration file
 ```
 
-Check out the [authentication page](authentication.md) for further information about the authentication means.
-
-You can manage multiple OVHcloud accounts using [profiles](profiles.md). Create a profile with `ovhcloud login --profile <name>`, switch between them with `ovhcloud config profile switch <name>`, or use `--profile <name>` on any command.
-
-### Generate Shell Completion
-
-```bash
-# Bash
-eval "$(./ovhcloud completion bash)"
-# Zsh
-eval "$(./ovhcloud completion zsh)"
-# Fish
-./ovhcloud completion fish | source
-# PowerShell
-./ovhcloud completion powershell | Out-String | Invoke-Expression
-```
-
-Add the appropriate line to your shell’s startup file (`~/.bashrc`, `~/.zshrc`, etc.) to enable persistent autocompletion.
-
----
-
-## Global Usage
-
-```text
-ovhcloud [command] [flags]
-```
-
-### Global Flags
-
-| Flag               | Description                                          |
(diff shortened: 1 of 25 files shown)


Denis-hamon:

**P4 · A command that spends money**  
_Dimension: Guard rail_

> **Expected:** The preview lists the eleven calls it would make and sends none. Ordering for real asks for the datacenter to be typed back, `--quote` stops at the price, and `--no-pay` places without paying. Is that enough friction for something that charges the account?

"This command can write, so live mode will not run it as is.

    baremetal order 24rise01-v3 --datacenter rbx --dry-run

If the command has no --dry-run, it will say so itself."

Denis-hamon:
P4 is green now, and the finding was in the review sandbox rather than in this branch.

The button fired `baremetal order 24rise01-v3 --datacenter rbx --dry-run`, and the sandbox's read-only guard rejected it before the binary ever ran — so the reviewer read the guard's refusal as the command's answer. The guard decides from the last component of the longest known command prefix, and it was checking the wrong token here.

Measured on the deployed build after the fix: `rc = 0`, and the preview lists the eleven calls it would make and sends none.

```
GET  /v1/me
GET  /v1/me/payment/method?default=true
POST /v1/order/cart
POST /v1/order/cart/{cartId}/assign
GET  /v1/order/cart/{cartId}/eco?planCode=24rise01-v3
POST /v1/order/cart/{cartId}/eco
GET  /v1/order/cart/{cartId}/item/{itemId}/requiredConfiguration
POST /v1/order/cart/{cartId}/item/{itemId}/configuration
GET  /v1/order/cart/{cartId}/checkout
POST /v1/order/cart/{cartId}/checkout
DELETE /v1/order/cart/{cartId}
```

That matches what the check expects. Nothing to change on this branch.

Denis-hamon:

**P4 · A command that spends money**  
_Dimension: Guard rail_

> **Expected:** The preview lists the eleven calls it would make and sends none. Ordering for real asks for the datacenter to be typed back, `--quote` stops at the price, and `--no-pay` places without paying. Is that enough friction for something that charges the account?

🔍 Dry run: nothing was sent. This would have been called:
  GET /v1/me
  GET /v1/me/payment/method?default=true
  POST /v1/order/cart
  POST /v1/order/cart/{cartId}/assign
  GET /v1/order/cart/{cartId}/eco?planCode=24rise01-v3
  POST /v1/order/cart/{cartId}/eco
  GET /v1/order/cart/{cartId}/item/{itemId}/requiredConfiguration
  POST /v1/order/cart/{cartId}/item/{itemId}/configuration
  GET /v1/order/cart/{cartId}/checkout
  POST /v1/order/cart/{cartId}/checkout
  DELETE /v1/order/cart/{cartId}
exit 0

PR #247, opened 2026-08-18
feat(order): follow up and settle what was ordered, instead of returning a number

Denis-hamon:
`baremetal order` (#246) places an order and hands back a number. Without these commands that number is a dead end: whether it was paid, where it is in delivery, and how to settle one placed with `--no-pay` all still require a browser — the exact break this audit exists to close.

Stacked on **#246**, itself stacked on **#245**. Review those first.

## Five commands, four read-only

```
ovhcloud order list [--days 30] [--limit 25] [--from T] [--to T]
ovhcloud order get <id>
ovhcloud order follow <id>
ovhcloud order pay <id> [--payment-method N]
ovhcloud order waive-retraction <id>
```

```
┌───────────┬────────────┬───────────┬──────────────┬────────────┐
│  orderId  │    date    │  status   │ priceWithTax │ retraction │
├───────────┼────────────┼───────────┼──────────────┼────────────┤
│ 256729207 │ 2026-08-16 │ delivered │ 5.48 €       │ kept       │
│ 256700494 │ 2026-08-14 │ delivered │ 0 PTS        │ kept       │
└───────────┴────────────┴───────────┴──────────────┴────────────┘
```

## An access token in every order

`GET /me/order/{id}` returns a `password`, and the same token again inside `url` and `pdfUrl`: the order form is reachable by **anyone holding the link**.

- `get` hands the link over — that is what was asked for — and says what it carries.
- `list` prints **no link at all**. Handing over the one order somebody asked about is a different act from printing fifty tokens for orders nobody did.
- The password is never printed under its own name, which is the string a log scanner greps for, and would be a second copy of something already shown.

Two tests guard this, and both were checked by making the leak happen.

## Bounds that are said, not silently applied

The state of an order is a **call of its own** — `GET /me/order/{id}` carries the price and the dates and no state — so a useful list costs two requests per order. On an account with 1103 orders that is worth bounding:

```
🟠 49 older orders in this window are not shown. Raise --limit, or narrow --days.
```

The limit keeps the **newest**: order identifiers ascend with time, so taking the head of the API's own list would show the three oldest under a heading that says recent.

## `follow` prints the status too, because the steps disagree with it

On a real order the API calls `delivered`, all four steps read `TODO` with no history:

```
Order 256729207 is delivered.
┌────────────┬────────┬────────┐
│    step    │ status │ latest │
├────────────┼────────┼────────┤
│ VALIDATING │ TODO   │        │
│ VALIDATED  │ TODO   │        │
│ DELIVERING │ TODO   │        │
│ AVAILABLE  │ TODO   │        │
└────────────┴────────┴────────┘
```

The table alone says nothing has happened.

## Two guards, deliberately unequal

**`waive-retraction` is a command precisely so that it is never a side effect of buying.** #246 refuses to send `waiveRetractationPeriod` even though the checkout body accepts it; giving up a legal right belongs in a sentence somebody typed on purpose. It cannot be undone, so it asks for the order number to be typed back.

**`pay` asks for a yes** — the order was agreed to once already, at a price that has not changed since. It also reads the state first and refuses an order that is not awaiting payment, rather than letting the API answer something confusing about a state the CLI had just read.

## Tests

Fifteen, on the API's own recorded answers — including the delivered order whose four steps all read `TODO`.

Each was checked by breaking what it guards: leaking the token into `list`, re-exposing the password in `get`, dropping the state lookup, paying and waiving without a confirmation, paying an order already settled, waiving twice. **Seven sabotages, seven red tests.**

## Verified against a real account

`list`, `get`, `follow`, and both write commands under `--dry-run`, on a real account with 1103 orders. No order was paid and no retraction period was waived.

<sub>Part of the BareMetal CLI UX audit (LVL2-16355), with #246.</sub>

### DIFF
--- a/doc/ovhcloud.md
+++ b/doc/ovhcloud.md
@@ -1,90 +1,36 @@
-# OVHcloud CLI (`ovhcloud`) Documentation
+## ovhcloud
 
----
+CLI to manage your OVHcloud services
 
-## Overview
+### Options
 
-`ovhcloud` is a single, unified command‑line interface for managing the full range of OVHcloud products and account resources directly from your terminal. Whether you need to automate provisioning, perform quick look‑ups, or integrate OVHcloud operations into CI/CD pipelines, `ovhcloud` offers fine‑grained commands and consistent output formats (table, JSON, YAML, or custom gval expressions).
-
----
-
-## Quick Start
-
-```bash
-# Display the top‑level help
-ovhcloud --help
-
-# Log in and create API credentials (interactive)
-ovhcloud login
-
-# List your VPS instances as JSON
-ohvcloud vps list -o json
+```
+  -d, --debug            Activate debug mode (will log all HTTP requests details)
+  -h, --help             help for ovhcloud
+  -e, --ignore-errors    Ignore errors in API calls when it is not fatal to the execution
+  -o, --output string    Output format: json, yaml, interactive, or a custom format expression (using https://github.com/PaesslerAG/gval syntax)
+                         Examples:
+                           --output json
+                           --output yaml
+                           --output interactive
+                           --output 'id' (to extract a single field)
+                           --output 'nested.field.subfield' (to extract a nested field)
+                           --output '[id, "name"]' (to extract multiple fields as an array)
+                           --output '{"newKey": oldKey, "otherKey": nested.field}' (to extract and rename fields in an object)
+                           --output 'name+","+type' (to extract and concatenate fields in a string)
+                           --output '(nbFieldA + nbFieldB) * 10' (to compute values from numeric fields)
+      --profile string   Use a specific profile from the configuration file
 ```
 
-Check out the [authentication page](authentication.md) for further information about the authentication means.
-
-You can manage multiple OVHcloud accounts using [profiles](profiles.md). Create a profile with `ovhcloud login --profile <name>`, switch between them with `ovhcloud config profile switch <name>`, or use `--profile <name>` on any command.
-
-### Generate Shell Completion
-
-```bash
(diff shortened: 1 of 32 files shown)


Denis-hamon:

**P5 · Following an order without leaving the terminal**  
_Dimension: Discoverability_

> **Expected:** You can answer “where is my server?” without opening the Manager. The states are the ones the Manager shows, and a delivered order says so plainly.

" This command can write, so live mode will not run it as is.

    order follow 256620146 --dry-run"

Denis-hamon:
P5 is green now, and the finding was in the review sandbox, not here.

`order follow` reads a delivery — it writes nothing — but it was missing from the sandbox's read-only whitelist, so the guard demanded `--dry-run` for a command that has none. The reviewer got the guard's refusal instead of the delivery status. A verb missing from that list breaks nothing visible: it just makes a check unplayable while dressing a read up as a write.

Measured on the deployed build after the fix, on an order that is still delivering:

```
┌────────────┬────────┬─────────────┐
│    step    │ status │   latest    │
├────────────┼────────┼─────────────┤
│ VALIDATING │ DONE   │ 2026-08-10  │
│ VALIDATED  │ DOING  │             │
│ DELIVERING │ DOING  │ 2026-08-10  │
│ AVAILABLE  │ TODO   │             │
└────────────┴────────┴─────────────┘
Order 256620146 is delivering.
```

Nothing to change on this branch.

Denis-hamon:
### Human review — Validated

All 2 checks of this pull request passed in the OVHcloud CLI review sandbox (reviewers: Denis, binary 70c936fa39ee).

PR #248, opened 2026-08-18
feat(baremetal): add power off/on/status, and say what powering off leaves behind

Denis-hamon:
Stacked on #245 — **only the last commit (`7e9fdfd`) belongs to this PR**; the three below it are the catalog work under review there.

## The API has no power endpoint. It has the verb anyway.

There is no `/power`, no `/shutdown`, no `/poweroff` in the 126 paths of `dedicated.server`, which is why "power a server off" reads as impossible from the CLI. It is spelled differently: `dedicated.server.BootTypeEnum` carries `power` as a first-class boot type, all 35 servers reachable from my account have an entry of that type (95083 in Europe, 95644 in Canada), and `bootId` is writable.

```
PUT  /dedicated/server/{n}  {"bootId": 95083}
POST /dedicated/server/{n}/reboot
```

The machine reboots into an entry whose only job is to shut it down.

```
ovhcloud baremetal power off <server> [--wait] [--timeout 10m] [--yes | --dry-run]
ovhcloud baremetal power on  <server> [--wait] [--timeout 10m] [--boot N] [--dry-run]
ovhcloud baremetal power status <server>
```

## Measured on real hardware, twice

First with a raw script, then through the CLI itself, on a spare RISE-2 with no OS installed. Starting state read before anything was touched, and returned to it:

```
$ ovhcloud baremetal power off ns30704xx… --yes --wait
✅ ns30704xx… is off.                                              (2 min 15 s)

$ ovhcloud baremetal power status ns30704xx…
ns30704xx… is poweroff, set to boot on "Power-off server" (bootId 95083)
⚠️  This server is set to power off: any reboot, including one from the
    manager, will shut it down again.

$ ovhcloud baremetal power on ns30704xx… --wait
✅ ns30704xx… is on, booting on 230242 (the boot it was on before it
   was powered off).                                               (4 min)
```

## Three measurements that shaped the code

**A `done` task does not mean the machine obeyed.** The `hardReboot` task reported `done` at 60 s; `powerState` only flipped at **t+143 s**, and at t+207 s on the way back up. So `--wait` polls `powerState` and never the task. Waiting on the task would have shipped a command that reports success while the server is still running.

**Powering off rewrites the boot configuration, durably.** The server stays on the power-off entry and shuts down again at every later reboot — including one triggered from the manager, with nothing on screen explaining why. Hence: `power off` records the previous boot, `power on` puts it back, and `power status` warns whenever a server sits on a power-off entry.

**The previous boot is not guessable.** My test server was in rescue (230242), not on disk — I had assumed disk, and restoring the assumption would have put the machine in a state it had never been in. The value is read before it is changed; when this machine has no record of it (a `power on` run from a different laptop than the `power off`), the command falls back to the disk entry **and says so in the message** rather than presenting a guess as a memory.

## One defect this surfaced in `internal/cache`

`cache.Write` calls `PurgeExpired(namespace, ttl)` with the ttl it is handed, so writing an empty entry with a zero ttl deletes **every other entry in the namespace**. Forgetting one server therefore forgot where all the others boot — while they were still powered off, which is exactly when that memory is the only thing between them and a boot they were never on. Fixed with a `cache.Remove`, with a test for the two-server case.

This is worth knowing beyond this PR: `cache.Write(ns, key, nil, 0)` is not a way to delete an entry.

## Notes

- `power on` takes `--dry-run` but no `--yes`: it interrupts nothing, so there is no prompt to skip. Caught by reading the generated docs, not the code.
- Ten tests, on responses recorded from the live API. Each was sabotaged and watched fail before being kept.
- `doc/ovhcloud_logout.md` is generated by `docgen` and has never been committed on `main`; it is untracked, so `git diff --exit-code` in the doc gate does not see it. Left out of this PR — it is not this change's business, but the gate has a hole.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/doc/ovhcloud.md
+++ b/doc/ovhcloud.md
@@ -1,90 +1,36 @@
-# OVHcloud CLI (`ovhcloud`) Documentation
+## ovhcloud
 
----
+CLI to manage your OVHcloud services
 
-## Overview
+### Options
 
-`ovhcloud` is a single, unified command‑line interface for managing the full range of OVHcloud products and account resources directly from your terminal. Whether you need to automate provisioning, perform quick look‑ups, or integrate OVHcloud operations into CI/CD pipelines, `ovhcloud` offers fine‑grained commands and consistent output formats (table, JSON, YAML, or custom gval expressions).
-
----
-
-## Quick Start
-
-```bash
-# Display the top‑level help
-ovhcloud --help
-
-# Log in and create API credentials (interactive)
-ovhcloud login
-
-# List your VPS instances as JSON
-ohvcloud vps list -o json
+```
+  -d, --debug            Activate debug mode (will log all HTTP requests details)
+  -h, --help             help for ovhcloud
+  -e, --ignore-errors    Ignore errors in API calls when it is not fatal to the execution
+  -o, --output string    Output format: json, yaml, interactive, or a custom format expression (using https://github.com/PaesslerAG/gval syntax)
+                         Examples:
+                           --output json
+                           --output yaml
+                           --output interactive
+                           --output 'id' (to extract a single field)
+                           --output 'nested.field.subfield' (to extract a nested field)
+                           --output '[id, "name"]' (to extract multiple fields as an array)
+                           --output '{"newKey": oldKey, "otherKey": nested.field}' (to extract and rename fields in an object)
+                           --output 'name+","+type' (to extract and concatenate fields in a string)
+                           --output '(nbFieldA + nbFieldB) * 10' (to compute values from numeric fields)
+      --profile string   Use a specific profile from the configuration file
 ```
 
-Check out the [authentication page](authentication.md) for further information about the authentication means.
-
-You can manage multiple OVHcloud accounts using [profiles](profiles.md). Create a profile with `ovhcloud login --profile <name>`, switch between them with `ovhcloud config profile switch <name>`, or use `--profile <name>` on any command.
-
-### Generate Shell Completion
-
-```bash
-# Bash
-eval "$(./ovhcloud completion bash)"
-# Zsh
-eval "$(./ovhcloud completion zsh)"
-# Fish
-./ovhcloud completion fish | source
-# PowerShell
-./ovhcloud completion powershell | Out-String | Invoke-Expression
-```
-
-Add the appropriate line to your shell’s startup file (`~/.bashrc`, `~/.zshrc`, etc.) to enable persistent autocompletion.
-
----
-
-## Global Usage
-
-```text
-ovhcloud [command] [flags]
-```
-
-### Global Flags
-
-| Flag               | Description                                          |
-| ------------------ | ---------------------------------------------------- |
-| `--debug`          | Activate debug mode (logs all HTTP‑request details). |
-| `--ignore-errors`  | Ignore errors of API calls made when listing items.  |
-| `--filter <expr>`  | Filter lists output with a [gval] expression.        |
-| `-h`, `--help`     | Display help for `ovhcloud` or a specific command.   |
-| `-o interactive`   | Produce interactive (prompt‑based) output.           |
-| `-o json`          | Output data in JSON format.                          |
-| `-o yaml`          | Output data in YAML format.                          |
-| `-o <expr>`        | Format output with a [gval] expression.              |
-
-[gval]: https://github.com/PaesslerAG/gval
-
-#### Filtering examples
-
-- Strict string equality: `--filter 'name=="something"'`
-- String regexp comparison: `--filter 'name=~"something"'`
-- Number comparison: `--filter 'bootId > 1'`
-
-#### Formatting example
-
-- Extract only one field: `-o 'ip'`
-- Extract an object: `-o '{name: ip}'`
-
----
-
-## Command Reference
-
-Below is the full list of primary sub‑commands available at the time of writing. Each can be explored in depth with `ovhcloud <command> --help`.
+### SEE ALSO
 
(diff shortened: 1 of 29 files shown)


Denis-hamon:

**B2 · Switching off a production machine**  
_Dimension: Guard rail_

> **Expected:** A confirmation prompt, the machine named, and it is stated that the machine will not come back on its own.

The output : 
🛑 failed to read the state of ns0000001.ip-203-0-113.eu: OVHcloud API error (status code 401): Client::Unauthorized: "You must login first" (X-OVH-Query-Id: EU.ext-3.6a8975b6.1312250.fc780d3974b76dec0eb4cbece12b1ac3)

Denis-hamon:

**B4 · Nothing to do: the command says so instead of acting**  
_Dimension: Product fit_

> **Expected:** No write call is sent, and the command says “nothing to do” while naming the entry.

This command can write, so live mode will not run it as is.

    baremetal power on ns0000002.ip-203-0-113.eu --dry-run

If the command has no --dry-run, it will say so itself.

Denis-hamon:
Followed up on today's two "Does not work" findings from the human review (B2, B4) — both trace back to the same cause, not to this PR's code.

**B2** (`power off`, 401 "You must login first" while reading the machine's state) and the sandbox's own `account get` were failing identically at review time: the sandbox's `OVH_ACCESS_TOKEN` had died (it carries no `exp` claim, but a previous one lasted ~27h — this one was already past that). Confirmed by decoding it: 26.6h old at the moment of the 401.

**B4** (`power on --dry-run`, message "This command can write... Add --dry-run") is the sandbox's own generic guard (`freemode.py`), shown for any write command run without `--dry-run` — it's templated boilerplate, not something this binary prints.

Token refreshed today. Rebuilt this branch and replayed both live against the test/integration account, on `ns3118333.ip-51-68-100.eu` (the machine Denis cleared for unrestricted testing):

```
$ ovhcloud baremetal power off ns3118333.ip-51-68-100.eu --dry-run
🔍 Dry run: nothing was sent. This would have been called:
  PUT /v1/dedicated/server/ns3118333.ip-51-68-100.eu
  POST /v1/dedicated/server/ns3118333.ip-51-68-100.eu/reboot

$ ovhcloud baremetal power on ns3118333.ip-51-68-100.eu --dry-run
ns3118333.ip-51-68-100.eu is already running on boot 1 (the disk, because this
machine has no record of the previous boot). Nothing to do.
```

Both work as intended — B4 in particular is exactly the "nothing to do, and the entry is named" behaviour the check asked for. No code change needed here.

Denis-hamon:

**B3 · Powering on restores the previous boot entry**  
_Dimension: Product fit_

> **Expected:** The behaviour — “I hand the machine back the way you left it” — is what you expect. A defect found here: the remembered entry was overwritten by the power-off entry, so `power on` restored a “power off”.

🛑 failed to read the state of ns0000001.ip-203-0-113.eu: OVHcloud API error (status code 404): Client::NotFound: "This service does not exist" (X-OVH-Query-Id: EU.ext-4.6a89a9d1.1101808.34160f0a4a397999e8833d9b0b29811a)

Denis-hamon:

**B4 · Nothing to do: the command says so instead of acting**  
_Dimension: Product fit_

> **Expected:** No write call is sent, and the command says “nothing to do” while naming the entry.

This command can write, so live mode will not run it as is.

    baremetal power on ns0000002.ip-203-0-113.eu --dry-run

If the command has no --dry-run, it will say so itself.

Denis-hamon:
Both of today's 15:53 findings come from the review sandbox, not from this branch. Fixed on the sandbox side.

**B3 — 404 "This service does not exist".** The check's steps offer `ovhcloud baremetal power status ns0000001.ip-203-0-113.eu` as a clickable button, and `ns0000001.ip-203-0-113.eu` is a documentation-range placeholder (RFC 5737). Correct in the prose; wrong the moment the button is fired in live mode, where it hits a real account on which no such server exists. So the 404 was the page aiming at nothing — you clicked exactly what it offered.

The sandbox now rewrites, at startup, every command **live mode would actually run** to a real test machine (`ns31633811.ip-162-19-234.eu`, RISE-1, lim3), covering both a check's own command and the `` `ovhcloud …` `` commands inside its steps — each of which becomes its own button. Commands the guards refuse are left untouched: they only ever replay, and recordings are keyed by the exact string, so each rewrite also aliases the recording. Verified through the sandbox's own API:

```
$ baremetal power status ns31633811.ip-162-19-234.eu
ns31633811.ip-162-19-234.eu is poweron, set to boot on "Boot to disk" (bootId 1)
rc = 0
```

Which is what B3 asks for: it names the entry it is on, so you can see for yourself which one it took.

**B4 — the "add --dry-run" guard text, again.** `power on` writes, so live mode refuses it without `--dry-run`; the message is the sandbox's generic guard, not this binary's. With the flag, on a machine that is already up:

```
$ baremetal power on ns31633811.ip-162-19-234.eu --dry-run
ns31633811.ip-162-19-234.eu is already running on boot 1 (the disk, because this
machine has no record of the previous boot). Nothing to do.
rc = 0
```

No write call sent, the entry named, and phrased as a statement rather than an error — the three things B4 asks for. Note the check's own command still says `ns0000002…` in the prose because `power on` is replay-only, but its button now aims at the real machine.

Separately: the sandbox was running a binary built on 21/08 17:21 and has been rebuilt from the current integration stack, so what the page shows is what these branches actually produce.

Denis-hamon:
### Human review — Validated

All 5 checks of this pull request passed in the OVHcloud CLI review sandbox (reviewers: Denis, binary 70c936fa39ee).

PR #253, opened 2026-08-19
feat(baremetal): say what a reinstall is doing while it does it

Denis-hamon:
> Stacked on #250. Only the last commit belongs to this pull request.

## A reinstall said nothing for twenty minutes

`reinstall --wait` followed the task, and the task API answers whether the work
is finished and nothing else. So this was the whole of it, repeated every
thirty seconds:

```
Still waiting for task 559351769 to complete (status=doing)…
```

`install/status` has the missing half — one entry per step, with the running
one marked — and the CLI used it **nowhere**.

## What it does now

```
ovhcloud baremetal install-status <server>     # read it, from anywhere
ovhcloud baremetal reinstall <server> --wait   # say which step, while waiting
```

Measured on a real reinstall of an ADVANCE-1 (EPYC 4245P, rbx8) — **11 min 05 s
end to end, exit 0**:

```
13:17:47 step 5/21: Preparing disks for new Partitioning
13:18:18 step 10/21: Applying Partitioning
13:18:49 step 13/21: Downloading OS image
13:19:19 step 16/21: Checking Partitioning
13:19:50 step 19/21: Waiting for services to be up
```

and `install-status`, against that same running installation:

```
**step 5 of 21 — Preparing disks for new Partitioning**

  # | Status | Step
  1 | done   | Checking BIOS version
  …
  5 | doing  | Preparing disks for new Partitioning
  6 | todo   | Cleaning Partitioning
  …
 21 | todo   | Sending end of installation mail
```

## Two things the API does that the code has to know about

**Outside an installation the route answers 404**, with *"Server is not being
installed or reinstalled at the moment"* — verified on three servers. That is
an answer, not a breakdown, so it is reported as one.

It is also why this endpoint **cannot carry the verdict of a `--wait`**: it says
the same thing before an installation starts and after it ends. The task stays
the authority on whether the work is done; `install/status` only says where it
has got to, and a progress read that fails is an empty line rather than a
failed command.

**`elapsedTime` is not the age of the installation.** Sampled every 30 s through
one real run:

| time | elapsedTime | step |
| --- | --- | --- |
| 13:17:44 | `+42` | 5 |
| 13:18:14 | `+72` | 10 |
| 13:18:44 | `+102` | 13 |
| 13:19:15 | `+133` | 16 |
| 13:19:45 | **`-106`** | 19 |
| 13:20:15 | `-76` | 19 |
| 13:21:16 | `-15` | 19 |

It counts up at exactly one per second and never resets between steps, but its
[...]

PR #254, opened 2026-08-19
feat(baremetal): read the traffic graphs, without knowing a MAC address

Denis-hamon:
> Stacked on #253. Only the last commit belongs to this pull request.

## The CLI reached none of the traffic data

Three routes carry it, and two are dead ends:

| route | state |
| --- | --- |
| `/dedicated/server/{sn}/mrtg` | deprecated **2017-10-23**, deletion announced 2018-04-23 |
| `/dedicated/server/{sn}/vrack/{vrack}/mrtg` | deprecated **2017-10-23**, marked LEGACY |
| `/dedicated/server/{sn}/networkInterfaceController/{mac}/mrtg` | **live** |

Both deprecated ones name `/networkInterfaceController` as their replacement,
so only that one is wired here.

## The route wants a MAC address, and nothing could tell you one

No command lists the controllers of a server. `vni list` shows the *virtual*
interfaces, and their own MAC list appears only under `-o json`. Asking an
operator for a MAC they have no way to look up would be a command nobody can
run, so `traffic` resolves the server to its controllers and reads every one.

```
ovhcloud baremetal traffic <server> [--period daily]
                                    [--type traffic:download,traffic:upload]
                                    [--nic <mac>]
```

Both directions are read by default: *how much is this server doing* is a
question about both, and answering one of them silently answers half.

Periods and types are checked against the API enums before the call, so a typo
comes back with the list of what would have worked rather than as a 400.

## Numbers appear twice, on purpose

Measured against a real ADVANCE-1, on the day it was reinstalled:

```
nic                type              peak         average      latest
9c:6b:00:a9:c5:10  traffic:download  33.98 Mbps   233.84 kbps  46.92 kbps
9c:6b:00:a9:c5:10  traffic:upload    116.23 kbps  1.69 kbps    2.32 kbps
9c:6b:00:a9:ca:25  traffic:download  944.68 bps   774.36 bps   739.16 bps
9c:6b:00:a9:ca:25  traffic:upload    15.73 bps    0.17 bps     0 bps
```

That peak is the operating system image being downloaded during the reinstall.

The raw figures behind those cells are `33978092.769` and an average of
`0.16518435754189942` — true, and useless to read. So the table shows a scaled
rate, while `-o json` and `-o peak` keep the raw values:

```
$ ovhcloud baremetal traffic <server> -o peak
33978092.769
```

`-o json` also carries **the whole series**. A day is 173 samples, which no
table can show and every script wants; dropping it would make the CLI's output
poorer than the API it wraps.

Every unit this API returns is a rate — `bps`, `pps` and `eps`, all three seen
on a real server — so one scale serves all three: `33.98 Mbps`, `2.87 kpps`,
`0 eps`.

## Verification

`go build`, `go vet`, `make wasm`, `go test ./...`, docgen — all green.

10 tests. **7 sabotages, 7 red**, each on the test meant to catch it: the scale
removed, the series dropped, `latest` replaced by the peak, an empty series
reported as zero, only the first controller read, and each of the two enum
checks disabled.

One of those sabotages first reported *no test falls* — it disabled the error
branch of `controllersOf` rather than shortening its result, so it never
restricted anything. Re-applied where it belonged, the test fell. The zero came
from the instrument, not from the code.

## Left out deliberately

`GET/PUT /dedicated/server/{sn}/burst` is live and uncovered, but it answers
404 *"The requested object (burst) does not exist"* on **8 of the 8 servers**
checked. The read path could be shipped and would report that answer, but the
path that returns a burst configuration cannot be exercised on this account, so
there is no command for it here.

### DIFF
--- a/doc/ovhcloud.md
+++ b/doc/ovhcloud.md
@@ -1,90 +1,36 @@
-# OVHcloud CLI (`ovhcloud`) Documentation
+## ovhcloud
 
----
+CLI to manage your OVHcloud services
 
-## Overview
+### Options
 
-`ovhcloud` is a single, unified command‑line interface for managing the full range of OVHcloud products and account resources directly from your terminal. Whether you need to automate provisioning, perform quick look‑ups, or integrate OVHcloud operations into CI/CD pipelines, `ovhcloud` offers fine‑grained commands and consistent output formats (table, JSON, YAML, or custom gval expressions).
-
----
-
-## Quick Start
-
-```bash
-# Display the top‑level help
-ovhcloud --help
-
-# Log in and create API credentials (interactive)
-ovhcloud login
-
-# List your VPS instances as JSON
-ohvcloud vps list -o json
+```
+  -d, --debug            Activate debug mode (will log all HTTP requests details)
+  -h, --help             help for ovhcloud
+  -e, --ignore-errors    Ignore errors in API calls when it is not fatal to the execution
+  -o, --output string    Output format: json, yaml, interactive, or a custom format expression (using https://github.com/PaesslerAG/gval syntax)
+                         Examples:
+                           --output json
+                           --output yaml
+                           --output interactive
+                           --output 'id' (to extract a single field)
+                           --output 'nested.field.subfield' (to extract a nested field)
+                           --output '[id, "name"]' (to extract multiple fields as an array)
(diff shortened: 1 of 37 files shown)


Denis-hamon:
### Human review — Validated

All 3 checks of this pull request passed in the OVHcloud CLI review sandbox (reviewers: Denis, binary 70c936fa39ee).

PR #257, opened 2026-08-19
feat(ip): the last 29 routes — licences, registry, delegation and the exit door

Denis-hamon:
## What this adds

This closes the `ip` domain. It had 15 of its 81 routes; with #255 and #256 below it, all the live
ones are reachable.

```
ovhcloud ip campus
ovhcloud ip licenses <ip_block>                          # the eight licence routes, in one answer
ovhcloud ip delegation list|get|add|remove
ovhcloud ip ripe get|set
ovhcloud ip byoip aggregations|slices|aggregate|slice
ovhcloud ip migration-token get|create [--reveal]
ovhcloud ip change-org <ip_block> <organisation>
ovhcloud ip service list|get|edit|change-contact|terminate|confirm-termination
ovhcloud ip service service-info get|edit
```

29 routes, 22 commands. Stacked on #256; **only the last commit belongs to this pull request.**

## What the account said

**`GET /ip/{ip}/delegation` answers HTTP 500 on 492 of 537 blocks.** The 45 that answer 200 are
exactly the IPv6 /56 and /64 blocks — 27 and 18. The failures are every IPv4 mask (336 × /32,
18 × /30, 8 × /28, 1 × /26) and every IPv6 /128 (129). **No prefix falls on both sides.** A block
that does not exist answers a correct `404 This service does not exist`, so the 500 is not a
not-found; and the sibling bring-your-own-IP routes show the API can do this properly — they answer
`400 This IP is not part of the Bring you own IP product.` on all 537. The API summarises the route
as *"Reverse delegation on IPv6 subnet"*, so that sentence is attached to the failure when the block
is IPv4. **This is the third API defect this audit has turned up**, after `install/hardwareRaidSize`
(404 on a route badged "Stable production version") and `/ip/{ip}/phishing` (500 outside Europe,
in #256). Worth a look by the route's owner.

**Three live migration tokens were sitting on this account.** The token is a bearer credential:
whoever holds it, with the customer identifier printed beside it, can take the address. So
`migration-token get` prints a fingerprint — `jamE… (20 characters)` — and `--reveal` prints the
token. The substitution happens on the object rather than in the template, so `-o json` is covered
by the same decision as the table: a masking that only applies to human-readable output protects
nothing, because the pipeline is where the value would have been logged.

**The eight licence routes have no index**, so "does this address carry a licence" was eight
requests. `ip licenses` asks them once. Measured: `139.99.131.186/32` carries `plesk-ca-75831` and
`windows-ca-75830`, and everything else on the sample carries none.

**`ripe set` reads before it writes.** The API takes the whole `RipeInfos` object, so a request
carrying only `--description` would publish an empty netname — in the public registry, without
anybody asking. The record is read and sent back with the changed field replaced.

**`ip list` and `ip service list` stay two commands** because they are two resources: 537 blocks and
80 services. A block is what gets routed; a service is what gets renewed, has contacts and can be
terminated — and one of the eighty cannot, which `--filter 'canBeTerminated==false'` finds.

**One route is deliberately absent**: `POST /ip/{ip}/terminate`, deprecated on 2016-07-21 with
deletion announced for 2017-01-21 and an `x-replacement-route` pointing at
`/ip/service/{serviceName}/terminate` — which is what this wires.

## Verified against the API

| command | measured |
| --- | --- |
| `ip campus` | 15 sites, their datacenters and the registries each accepts |
| `ip licenses 139.99.131.186/32` | 2 licences · another block: none |
| `ip ripe get 151.80.69.32/30` | `OVH_251564576` / `Failover Ips` |
| `ip delegation list` on an IPv6 /64 | no delegation set |
| `ip delegation list` on an IPv4 /32 | the 500, with what the route covers |
| `ip service list --filter 'canBeTerminated==false'` | the one service of eighty |
| `ip service service-info get` | contacts, renewal, expiry |
| `ip migration-token get` | customer named, token withheld — `-o json` too |
| `ip migration-token get --reveal` | the token |
| `ip byoip aggregations` / `slices` | the API's own "not part of the Bring you own IP product" |

Ten refusals executed against the API: an IPv4 delegation, both byoip previews, a `ripe set` with
nothing to set, a token creation without a customer, a contact change with no contact, an empty
termination token, an unknown termination reason, and both byoip writes without their required
value.

Ten writes proven by `--dry-run`, including the termination preview, which names every field it
would send and fingerprints the token: `token: Supe… (18 characters), reason: TOO_EXPENSIVE`.
Nothing was written — the addresses carrying game rules, mitigation and RIPE records belong to
running services.

## Notes

- The termination helpers — the survey enum check, the completion, the token fingerprint — move to
  `internal/services/common` in the commit below this one. They were written for `baremetal` in
  #244; the enums are shared API types, and a second copy of a routine that withholds a credential
  does not expire, it stays wrong in silence while the original is fixed. The six existing baremetal
  termination tests are the regression net for the move and pass unchanged.
- The accepted termination reasons are read from the specification embedded in the binary rather
  than transcribed: fourteen values today, and a list copied into Go drifts the day the API gains a
  fifteenth — silently, into a 400 nobody can explain.

---

## Review round — second commit

An adversarial review of the first commit found five defects, all real.

**The delegation note covered IPv4 and not the IPv6 /128** — the one case its own comment names,
quoting the measurement that says the 500 hits every IPv4 mask *and* every /128. A single address
got a bare 500 and no explanation. `delegation get` had it worse: it went through the generic
object helper and never reached the explanation at all.

**Three guards were bypassable.** `--size -1` satisfied "a value was given" and travelled to the
API as a prefix length, through a confirmation offering to slice into `/-1` blocks. `--netname "  "`
satisfied "something to change" and would have published a blank netname to the public registry,
behind a yes/no. `--customer-id "  "` did the same on a token creation. The termination token was
already trimmed before being tested; these three were not.

**`byoip aggregate` and `byoip slice` reported a failed *read*** after a refused write, because they
shared the prefix of the two preview routes.

**`migration-token get` blamed a missing token for every failure**, sending the operator to
`migration-token create` after a 403 or a 500 — which fails for the reason that message just hid.

Seven sabotages against the fixes, seven red. The five refusals were also run against the API,
including the `/128` that now says what it is:

```
🛑 failed to read the reverse delegation of 2001:…::77b/128: … (status code 500) …
   This route covers reverse delegation on IPv6 subnets; 2001:…::77b/128 is a single IPv6 address.
```

### DIFF
--- a/doc/ovhcloud.md
+++ b/doc/ovhcloud.md
@@ -1,90 +1,36 @@
-# OVHcloud CLI (`ovhcloud`) Documentation
+## ovhcloud
 
----
+CLI to manage your OVHcloud services
 
-## Overview
+### Options
 
-`ovhcloud` is a single, unified command‑line interface for managing the full range of OVHcloud products and account resources directly from your terminal. Whether you need to automate provisioning, perform quick look‑ups, or integrate OVHcloud operations into CI/CD pipelines, `ovhcloud` offers fine‑grained commands and consistent output formats (table, JSON, YAML, or custom gval expressions).
-
----
-
-## Quick Start
-
-```bash
-# Display the top‑level help
-ovhcloud --help
-
-# Log in and create API credentials (interactive)
-ovhcloud login
-
-# List your VPS instances as JSON
-ohvcloud vps list -o json
+```
+  -d, --debug            Activate debug mode (will log all HTTP requests details)
+  -h, --help             help for ovhcloud
+  -e, --ignore-errors    Ignore errors in API calls when it is not fatal to the execution
+  -o, --output string    Output format: json, yaml, interactive, or a custom format expression (using https://github.com/PaesslerAG/gval syntax)
+                         Examples:
+                           --output json
+                           --output yaml
+                           --output interactive
+                           --output 'id' (to extract a single field)
+                           --output 'nested.field.subfield' (to extract a nested field)
+                           --output '[id, "name"]' (to extract multiple fields as an array)
+                           --output '{"newKey": oldKey, "otherKey": nested.field}' (to extract and rename fields in an object)
+                           --output 'name+","+type' (to extract and concatenate fields in a string)
+                           --output '(nbFieldA + nbFieldB) * 10' (to compute values from numeric fields)
+      --profile string   Use a specific profile from the configuration file
 ```
 
-Check out the [authentication page](authentication.md) for further information about the authentication means.
-
-You can manage multiple OVHcloud accounts using [profiles](profiles.md). Create a profile with `ovhcloud login --profile <name>`, switch between them with `ovhcloud config profile switch <name>`, or use `--profile <name>` on any command.
-
-### Generate Shell Completion
-
-```bash
-# Bash
-eval "$(./ovhcloud completion bash)"
-# Zsh
-eval "$(./ovhcloud completion zsh)"
-# Fish
-./ovhcloud completion fish | source
-# PowerShell
-./ovhcloud completion powershell | Out-String | Invoke-Expression
-```
-
-Add the appropriate line to your shell’s startup file (`~/.bashrc`, `~/.zshrc`, etc.) to enable persistent autocompletion.
-
----
-
-## Global Usage
-
-```text
-ovhcloud [command] [flags]
-```
-
-### Global Flags
-
-| Flag               | Description                                          |
-| ------------------ | ---------------------------------------------------- |
-| `--debug`          | Activate debug mode (logs all HTTP‑request details). |
-| `--ignore-errors`  | Ignore errors of API calls made when listing items.  |
-| `--filter <expr>`  | Filter lists output with a [gval] expression.        |
-| `-h`, `--help`     | Display help for `ovhcloud` or a specific command.   |
-| `-o interactive`   | Produce interactive (prompt‑based) output.           |
-| `-o json`          | Output data in JSON format.                          |
-| `-o yaml`          | Output data in YAML format.                          |
-| `-o <expr>`        | Format output with a [gval] expression.              |
-
-[gval]: https://github.com/PaesslerAG/gval
-
-#### Filtering examples
-
-- Strict string equality: `--filter 'name=="something"'`
-- String regexp comparison: `--filter 'name=~"something"'`
-- Number comparison: `--filter 'bootId > 1'`
-
-#### Formatting example
-
-- Extract only one field: `-o 'ip'`
-- Extract an object: `-o '{name: ip}'`
-
----
-
-## Command Reference
-
-Below is the full list of primary sub‑commands available at the time of writing. Each can be explored in depth with `ovhcloud <command> --help`.
+### SEE ALSO
 
 * [ovhcloud account](ovhcloud_account.md)	 - Manage your account
 * [ovhcloud alldom](ovhcloud_alldom.md)	 - Retrieve information and manage your AllDom services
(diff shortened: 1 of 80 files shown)


Denis-hamon:

**G1 · A migration token is a bearer secret**  
_Dimension: Secrets_

> **Expected:** A fingerprint is shown, not the token. `--reveal` shows it, and the substitution happens on the object — so `-o json` is masked too.

🛑 no migration token exists for 162.19.234.217/32: OVHcloud API error (status code 404): "The requested object (migrationToken) does not exist" (X-OVH-Query-Id: EU.ext-5.6a8b473c.1090048.2fb32ca664b8709a7693d9287ca3b0a5)
   Create one with: ovhcloud ip migration-token create 162.19.234.217/32 --customer-id <customer>
exit 1

Denis-hamon:
Follow-up on G1 and G3.

**G1 was a false negative, and it belonged to the review sandbox.** The check ran against the IP the sandbox retargets everything to, and that address has no pending migration token — hence the 404. Measured across the account on 24/08: **3 of 537 IPs carry a token**. Pointing the check at one of them shows exactly what it asks for:

```
 Key                                 | Value
-------------------------------------|------------------------------------
 Customer                            | ts29436-ovh
 Token                               | 2pDY… (20 characters)

💡 The token is withheld. Print it with --reveal — whoever holds it can claim this IP.
```

The command is correct: fingerprint by default, `--reveal` to print, substitution on the object so `-o json` is masked too. The sandbox now retargets **per command** rather than sending everything to one machine — a fictitious identifier does not have one real equivalent, it has one per command — and a check whose object exists nowhere on the account is now announced as replay-only, with the reason, instead of returning a 404 the reviewer charges to the CLI.

**G3 cannot be judged from the button, and the sandbox now says so before the click.** The guard turns every write into `--dry-run`, and a dry run has no prompt — which is the entire subject of that check. It is judged on the recording.

Denis-hamon:
### Human review — Doubt

**G3 · Changing the holder of an IP**  
_Dimension: Guard rail_

> **Expected:** Treated as destructive: an explicit prompt, and it is stated that the IP leaves the account.

🔍 Dry run: nothing was sent.
  It would have asked first: This hands 162.19.234.217/32 over to the organisation ORG-EXAMPLE-1: the address leaves this account, and the change is published to the regional registry. Taking it back needs the new holder to agree.
This would have been called:
  POST /v1/ip/162.19.234.217%2F32/changeOrg
exit 0

Denis-hamon:

**G1 · A migration token is a bearer secret**  
_Dimension: Secrets_

> **Expected:** A fingerprint is shown, not the token. `--reveal` shows it, and the substitution happens on the object — so `-o json` is masked too.

# 🎫 Migration token for 46.105.33.152/32                                   
                                                                              
   Key                                 | Value                                
  -------------------------------------|------------------------------------  
   Customer                            | ts29436-ovh                          
   Token                               | 2pDY… (20 characters)                
                                                                              
  💡 The token is withheld. Print it with --reveal — whoever holds it can     
  claim                                                                       
  this IP.

Denis-hamon:

**G1 · A migration token is a bearer secret**  
_Dimension: Secrets_

> **Expected:** A fingerprint is shown, not the token. `--reveal` shows it, and the substitution happens on the object — so `-o json` is masked too.

# 🎫 Migration token for 46.105.33.152/32                                   
                                                                              
   Key                                 | Value                                
  -------------------------------------|------------------------------------  
   Customer                            | ts29436-ovh                          
   Token                               | 2pDY… (20 characters)                
                                                                              
  💡 The token is withheld. Print it with --reveal — whoever holds it can claim this IP.

PR #261, opened 2026-08-19
feat(support): open a ticket that already knows which machine it is about

Denis-hamon:
Stacked on #259. **Only the last commit belongs to this pull request.**

C8 of the audit asks for support-ticket creation "pre-filled with machine context". The support domain turned out to be **already wired** — `list`, `get`, `messages`, `reply`, `create` — so this is not a coverage lot. What was missing is that nothing joined a ticket to the server it is about: `create` has accepted a `serviceName` all along, and the operator still retyped the server name and described its state from memory.

## What this adds

```
ovhcloud baremetal ticket <server> --subject '...' --body '...' [--no-context]

ovhcloud support-tickets close <id>
ovhcloud support-tickets reopen <id> --reason '...'
ovhcloud support-tickets can-be-scored <id>
ovhcloud support-tickets list --status open --service-name ns1.example ...
```

`baremetal ticket` fills `product` and `serviceName`, and appends what the machine *is*: commercial range, datacenter, active boot, monitoring, running tasks, planned maintenance, and whatever `baremetal doctor` reports about it. The description stays the operator's — the collected block says what the machine is, not what is wrong.

**A server that cannot be read is written into the ticket rather than stopping it.** That is the situation somebody opens a ticket for; a support agent learns more from "the CLI could not read this server, here is the error" than from a ticket that was never sent.

## What the specification says, measured

The embedded schema and the live one (`eu.api.ovh.com/1.0/support.json`) were compared route by route: **identical, 9 routes on both sides.** No staleness this time, unlike the game protocols in #256.

Five of the nine routes had commands. Three of the remaining four are added here.

**`canBeClosed` is a field, `canBeScored` is a route.** The same question asked twice, by two mechanisms. So `close` checks without an extra request, using the read it already makes to name the ticket in its prompt.

**Setting the score is deliberately absent.** The API types `score` as a free string in both the embedded and the live specification, with no enum on either side. Establishing the accepted values would mean sending real feedback about a real support agent. Same reasoning as `burst` in #254.

## The list filtered nothing server-side

`GET /support/tickets` accepts eleven filters and none was reachable. `--filter` works at the other end of the wire: it drops rows the CLI has already fetched, and this collection returns identifiers that are expanded one request per ticket.

**The filters belong to the collection call and to nothing else.** The expansion builds one URL per ticket by appending to the path, so a query carried over would request `/support/tickets?status=open/12345` — a path the API cannot route, for a filter that would have looked like it worked. A test pins that, and a sabotage that reattaches it turns red.

**The three booleans are only sent when they were typed.** All three are meaningful in both positions — `archived=false` asks for the tickets that are *not* archived — so a flag left alone cannot be encoded as false without inventing a filter the operator never asked for.

Noted, not fixed here: `FetchArray` follows `X-Pagination-Cursor-Next`, which is the v2 mechanism, while this v1 collection paginates by `page`/`pageSize` in the query. The CLI sends neither and takes whatever the API defaults to. Establishing whether that truncates needs a live call against an account with enough tickets, which this branch has not been able to make.

## Verification

`go build`, `go vet`, `make wasm`, `go test ./...`, docgen — all green. **9 unit tests, 19 cobra tests, 12 sabotages, 12 red**, restoration verified by checksum on the three touched files.

Every command named in a message was checked against the real command tree of the built binary, not assumed.

**The write paths are proven by `--dry-run`, unit tests and cobra tests against recorded responses — not against the live API.** The credentials available to this branch expired before the sequence could run. The sequence itself is written and unchanged: create a ticket on a test server, read it back, verify out of the CLI that the body carries the machine context, `can-be-scored`, close, verify, refuse a second close, reopen, refuse a blank reason, close again, verify the ticket is back at rest. It will be run and this section replaced with what it measured.


Denis-hamon:

**J1 · `baremetal ticket` carries the machine context**  
_Dimension: Discoverability_

> **Expected:** The ticket body is pre-filled with the state read from the machine.

This command can write, so live mode will not run it as is.

    baremetal ticket ns0000001.ip-203-0-113.eu --subject disk failure --dry-run

If the command has no --dry-run, it will say so itself.
exit 77

Denis-hamon:
Followed up on today's "Does not work" finding (J1) — the sandbox's generic write guard ("This command can write... Add --dry-run") fired because the check was run without `--dry-run`, not because of anything specific to this command.

Rebuilt this branch and replayed live against the test/integration account:

```
$ ovhcloud baremetal ticket ns3118333.ip-51-68-100.eu --subject "disk failure" \
    --body "..." --dry-run
🔍 Dry run: nothing was sent. This would have been created:

Subject: disk failure
Product: dedicated
Service: ns3118333.ip-51-68-100.eu

...

--- Collected automatically by the OVHcloud CLI (ovhcloud baremetal ticket) ---
Service name         ns3118333.ip-51-68-100.eu
Commercial range     ADVANCE-1 | AMD EPYC 4245P
Datacenter           rbx8
...
Active boot          #1 harddisk, kernel hd
Running tasks        2
...
```

The machine's state is pre-filled exactly as J1 expects. No code change needed here.

Denis-hamon:
### Human review — Doubt

**J1 · `baremetal ticket` carries the machine context**  
_Dimension: Discoverability_

> **Expected:** The ticket body is pre-filled with the state read from the machine.

🔍 Dry run: nothing was sent. This would have been created:

Subject: disk failure
Product: dedicated
Service: ns31633811.ip-162-19-234.eu

The second disk reports read errors

--- Collected automatically by the OVHcloud CLI (ovhcloud baremetal ticket) ---
Service name         ns31633811.ip-162-19-234.eu
Commercial range     RISE-1 | Intel Xeon-E 2386G
Datacenter           lim3
Rack                 LIM0311A03C
Operating system     ubuntu2404-server_64
Main IP              162.19.234.217
Reverse              ns31633811.ip-162-19-234.eu
State                ok
Power state          poweron
Support level        pro
Professional use     false
Monitoring           true
No intervention      false
Active boot          #1 harddisk, kernel hd
Running tasks        0
Planned maintenance  0

Checks from ovhcloud baremetal doctor, except its renewal check (not run here):
  nothing
exit 0

PR #283, opened 2026-09-07
feat(managed-database): add network update support to edit command

AchrafHammouda:
feat(managed-database): add network update support to edit command

# Description

Add `--network-id`, `--subnet-id`, and `--public-network` flags to the `managed-database edit` command, enabling in-place network switching between public and private networks.

Usage:
- Switch to private: `--network-id <uuid> --subnet-id <uuid>`
- Switch to public: `--public-network`

The `--public-network` flag is mutually exclusive with `--network-id` and `--subnet-id`.

Network fields are injected into the `EditResource` body via a new `extraFields` variadic parameter, bypassing the OpenAPI spec filter in a single PUT call.

Changes:
- Added `--network-id`, `--subnet-id`, and `--public-network` flags to the edit command
- Added `extraFields` variadic parameter to `common.EditResource` to inject fields after the OpenAPI filter
- Updated auto-generated documentation with network update examples
- Added usage examples in the command's help text

# How Has This Been Tested?

Manually tested against a live PostgreSQL production cluster on OVH EU:

- [x] Public to private network switch (`--network-id <uuid> --subnet-id <uuid>`) — cluster switched to private successfully
- [x] Private to public network switch (`--public-network`) — cluster switched to public successfully
- [x] Mutual exclusivity (`--public-network --network-id <uuid>`) — properly rejected by cobra
- [x] Edit without network flags while on public (`--description "..."`) — no network change triggered
- [x] Edit without network flags while on private (`--description "..."`) — no network change triggered

## Type of change

- [x] Documentation update

# Checklist:

- [x] I ran `go mod tidy`

### DIFF
--- a/internal/cmd/cloud_managed_database.go
+++ b/internal/cmd/cloud_managed_database.go
@@ -317,6 +317,21 @@ There are two ways to define the edition parameters:
   Note that it is also possible to override values in the presented examples using command line flags like the following:
 
 	ovhcloud cloud managed-database edit <service_id> --editor --description "My database cluster"
+
+Network update:
+
+  You can switch a database service between public and private networks without recreating it.
+
+  To switch from public to private network:
+
+	ovhcloud cloud managed-database edit <service_id> --network-id <private-network-uuid> --subnet-id <subnet-uuid>
+
+  To switch from private to public network:
+
+	ovhcloud cloud managed-database edit <service_id> --network-id none --subnet-id none
+
+  Note: Changing the network triggers a service rebuild. The service will be temporarily unavailable
+  during the transition.
 `,
 		ValidArgsFunction: completion.CloudResources("/v1/cloud/project/%s/database"),
 		Run:               cloud.EditManagedDatabase,
@@ -336,6 +351,8 @@ There are two ways to define the edition parameters:
 
 	// Network configuration
 	managedDatabaseEditCmd.Flags().StringSliceVar(&cloud.ManagedDatabaseSpec.CLIIPRestrictions, "ip-restrictions", nil, "IP blocks authorized to access the cluster (CIDR format)")
+	managedDatabaseEditCmd.Flags().StringVar(&cloud.ManagedDatabaseSpec.CLINetworkID, "network-id", "", `Private network ID (use "none" to switch to public network)`)
+	managedDatabaseEditCmd.Flags().StringVar(&cloud.ManagedDatabaseSpec.CLISubnetID, "subnet-id", "", `Private subnet ID (use "none" to switch to public network)`)
 
 	// Common flags for other mean to define parameters
 	addInteractiveEditorFlag(managedDatabaseEditCmd)
--- a/internal/services/cloud/cloud_managed_database.go
+++ b/internal/services/cloud/cloud_managed_database.go
@@ -87,6 +87,8 @@ var (
 		Engine            string   `json:"-"`
 		CLIIPRestrictions []string `json:"-"`
 		CLINodesList      []string `json:"-"`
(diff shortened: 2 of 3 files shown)


internal/services/cloud/cloud_managed_database.go
```diff
@@ -236,16 +238,44 @@ func EditManagedDatabase(cmd *cobra.Command, args []string) {
...
+			ManagedDatabaseSpec,
+			assets.CloudOpenapiSchema,
+		); err != nil {
+			display.OutputError(&flags.OutputFormatConfig, "%s", err)
+			return
+		}
+	}
+
+	// Send network update as a direct PUT
+	if networkChanged {
```
amstuta: could you do this before the `common.EditResource` call, so you avoid this second PUT call ?

internal/cmd/cloud_managed_database.go
```diff
@@ -317,6 +317,21 @@ There are two ways to define the edition parameters:
...
+
+  You can switch a database service between public and private networks without recreating it.
+
+  To switch from public to private network:
+
+	ovhcloud cloud managed-database edit <service_id> --network-id <private-network-uuid> --subnet-id <subnet-uuid>
+
+  To switch from private to public network:
+
+	ovhcloud cloud managed-database edit <service_id> --network-id none --subnet-id none
```
amstuta: this explicit `none` value doesn't feel like the rest of the CLI commands, could you replace it by another flag (something like `--public-network`) that would be mutually exclusive with the `--network-id` and `--subnet-id` flags instead ?

internal/services/cloud/cloud_managed_database.go
```diff
@@ -238,44 +239,32 @@ func EditManagedDatabase(cmd *cobra.Command, args []string) {
...
-			cmd,
-			fmt.Sprintf("/cloud/project/{serviceName}/database/%s/{clusterId}", url.PathEscape(databaseService["engine"].(string))),
-			endpoint,
-			ManagedDatabaseSpec,
-			assets.CloudOpenapiSchema,
-		); err != nil {
-			display.OutputError(&flags.OutputFormatConfig, "%s", err)
-			return
+	// Build extra network fields that bypass the OpenAPI filter
+	var networkFields map[string]any
```
amstuta, line 243: I don't get why you create an extra map of fields instead of using `ManagedDatabaseSpec.CLIPublicNetwork` to fill `ManagedDatabaseSpec.NetworkID` and `ManagedDatabaseSpec.SubnetID` ?

AchrafHammouda: The extraFields approach is needed for two reasons:

  1. OpenAPI filter strips networkId/subnetId — EditResource calls FilterEditableFields which removes any field not declared in the PUT schema. Even if we set ManagedDatabaseSpec.NetworkID and .SubnetID directly, they get filtered out before the PUT request is sent. The extraFields parameter injects them after the filter, in a single PUT call.
  2. null vs empty string — When switching to public network, the API expects "networkId": null. But NetworkID is a Go string with omitempty — an empty string is omitted from the JSON entirely, not serialized as null. The map[string]any with nil values is the only way to produce proper JSON null.

amstuta: For your point 1), I don't see why the API-accepted fields would be dropped before the request is sent, we already do this kind of thing here: https://github.com/ovh/ovhcloud-cli/blob/main/internal/services/cloud/cloud_managed_database.go#L235

Anyway, your solution works even if I'm not a big fan of the code, so I'll merge it.

amstuta (approved):

PR #284, opened 2026-09-09
Migrate snapshot file storage v2

Bxota:
# Description

Migrates the file storage share snapshot commands to APIv2
(`/v2/publicCloud/project/{projectId}/storage/file/snapshot`) and aligns the embedded
`cloud_v2.json` schema with the published APIv2 schema.

Snapshot routes are project scoped in v2 (they are no longer nested under a share), so:

- The parent share is now referenced through `targetSpec.share.id`, and `snapshot list`
  filters the project-wide listing on it.
- `get`, `edit` and `delete` check that the requested snapshot really belongs to the share
  given on the command line, so a snapshot cannot be read or deleted through an unrelated
  share ID.
- The listing columns only read `targetSpec` and the root `resourceStatus`, so a snapshot
  that is still being created — and therefore has no `currentState` yet — is listed like
  any other one instead of aborting the whole listing.
- A snapshot inherits the region of its parent share: `targetSpec` has no `location`, so the
  extra `GET` on the share that was only there to resolve the region is gone.

Schema alignment (`internal/assets/api-schemas/cloud_v2.json`):

- `FileStorageSnapshotCurrentState`: `snapshotSize` renamed to `size`, `shareProto` and
  `shareSize` removed (they are not part of the contract).
- `FileStorageSnapshotTargetSpec`: `location` removed, `share` is the only required field.
- Request/response examples updated accordingly.

Command changes:

- `snapshot create` now goes through the shared creation helper, so it supports `--from-file`,
  `--init-file` and `--editor`, validates the payload client-side, and accepts `--wait`.
  Creation is asynchronous, so the default output now reports that creation *started*.
- `snapshot edit <share_id> <snapshot_id>` is added (`PUT`, name and description), with
  `--editor` and `--wait`.
- `snapshot delete` reports that the snapshot *is being deleted*, matching its asynchronous
  behavior.
- The `--region` persistent flag is removed: it only fed the v1 region discovery, which no
  longer exists now that every file storage route is project scoped.

The detail template only referenced fields that do not exist in the API and crashed on a
snapshot whose `currentState` is still null (right after creation); it now renders the real
fields and degrades gracefully.

Also removes 17 stale `doc/ovhcloud_cloud_storage-file_*.md` files left over from the
`storage-file` -> `storage file` command rename, and regenerates `doc/`.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Breaking change (fix or feature that can break a current behavior)
- [x] Documentation update

Breaking change: the `--region` flag of `ovhcloud cloud storage file` is removed.

# Checklist:

- [x] I ran `go mod tidy`

### DIFF
--- a/doc/ovhcloud_cloud_storage_file_share_snapshot_edit.md
+++ b/doc/ovhcloud_cloud_storage_file_share_snapshot_edit.md
@@ -1,15 +1,19 @@
-## ovhcloud cloud storage file share delete
+## ovhcloud cloud storage file share snapshot edit
 
-Delete the given share
+Edit a snapshot of the given share
 
 ```
-ovhcloud cloud storage file share delete <share_id> [flags]
+ovhcloud cloud storage file share snapshot edit <share_id> <snapshot_id> [flags]
 ```
 
 ### Options
 
 ```
-  -h, --help   help for delete
+      --description string   Snapshot description
+      --editor               Use a text editor to define parameters
+  -h, --help                 help for edit
+      --name string          Snapshot name
+      --wait                 Wait for the snapshot to be ready before exiting
 ```
 
 ### Options inherited from parent commands
@@ -34,5 +38,5 @@ ovhcloud cloud storage file share delete <share_id> [flags]
 
 ### SEE ALSO
 
-* [ovhcloud cloud storage file share](ovhcloud_cloud_storage_file_share.md)	 - Manage file storage shares
+* [ovhcloud cloud storage file share snapshot](ovhcloud_cloud_storage_file_share_snapshot.md)	 - Manage share snapshots
 
--- a/internal/services/cloud/cloud_storage_file.go
+++ b/internal/services/cloud/cloud_storage_file.go
@@ -12,7 +12,6 @@ import (
 
 	"github.com/ovh/ovhcloud-cli/internal/assets"
 	"github.com/ovh/ovhcloud-cli/internal/display"
-	filtersLib "github.com/ovh/ovhcloud-cli/internal/filters"
 	"github.com/ovh/ovhcloud-cli/internal/flags"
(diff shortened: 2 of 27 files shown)


doc/ovhcloud_cloud_storage_file_share_snapshot_edit.md
```diff
@@ -0,0 +1,42 @@
+## ovhcloud cloud storage file share snapshot edit
+
+Edit a snapshot of the given share
+
+```
+ovhcloud cloud storage file share snapshot edit <share_id> <snapshot_id> [flags]
```
amstuta, line 6: the snapshots commands should be at the same level as `share` to mimic the API path (`/publicCloud/project/{projectId}/storage/file/snapshot`) => `ovhcloud cloud storage file snapshot create|edit|get|…`

internal/services/cloud/cloud_storage_file.go
```diff
@@ -376,71 +362,170 @@ func DeleteShareACL(_ *cobra.Command, args []string) {
...
 	if err != nil {
 		display.OutputError(&flags.OutputFormatConfig, "%s", err)
 		return
 	}
 
-	var snapshots []map[string]any
-	if err := httpLib.Client.Get(endpoint+"/snapshot", &snapshots); err != nil {
-		display.OutputError(&flags.OutputFormatConfig, "failed to fetch share snapshots: %s", err)
+	filters := append([]string{}, flags.GenericFilters...)
+	filters = append(filters, fmt.Sprintf("targetSpec.share.id==%q", args[0]))
```
amstuta, line 416: listing should be global by default (not filtered by parent share), and have an optional argument to add this filter

amstuta (changes requested):

